PR #4, opened 2018-04-29
Fixed months to be in english. Added make_index journal function

kwrobert:
I kind of missed the `:VimwikiDiaryGenerateLinks` function from Vimwiki, so I added something similar here. First shot at writing any vimscript, so if everything looks good I'll add a proper command for people to use.

I did have a question, which is about how one should deal with joining paths in vimscript in a platform-agnostic way. Is there a community-standard way of joining strings into a path that is platform-independent? Something like Python's `os.path.join`?

### DIFF
--- a/autoload/wiki/journal.vim
+++ b/autoload/wiki/journal.vim
@@ -1,0 +1,132 @@
+" A simple wiki plugin for Vim
+"
+" Maintainer: Karl Yngve Lervåg
+" Email:      karl.yngve@gmail.com
+" License:    MIT license
+"
+
+function! wiki#journal#make_note(...) abort " {{{1
+  let l:date = (a:0 > 0 ? a:1 : strftime('%Y-%m-%d'))
+  call wiki#url#parse('journal:' . l:date).open()
+endfunction
+
+" }}}1
+function! wiki#journal#copy_note() abort " {{{1
+  let l:next_day = wiki#date#get_next_weekday(expand('%:t:r'))
+
+  let l:next_entry = printf('%s/%s/%s.wiki',
+        \ wiki#get_root(), g:wiki_journal, l:next_day)
+  if !filereadable(l:next_entry)
+    execute 'write' l:next_entry
+  endif
+
+  call wiki#url#parse('journal:' . l:next_day).open()
+endfunction
+
+" }}}1
+function! wiki#journal#go(step) abort " {{{1
+  let l:links = s:get_links()
+  let l:index = index(l:links, expand('%:t:r'))
+  let l:target = l:index + a:step
+
+  if l:target >= len(l:links) || l:target <= 0
+    return
+  endif
+
+  call wiki#url#parse('journal:' . l:links[l:target]).open()
+endfunction
+
+" }}}1
+function! wiki#journal#go_to_week() abort " {{{1
+  let l:date = expand('%:r') =~# '\d\d\d\d-\d\d-\d\d'
+        \ ? expand('%:r')
+        \ : strftime('%F')
+  call wiki#url#parse('journal:' . l:date[:3]
+        \ . '_w' . wiki#date#get_week(l:date)).open()
+endfunction
+
+" }}}1
+function! wiki#journal#go_to_month() abort " {{{1
+  let l:date = expand('%:r') =~# '\d\d\d\d-\d\d-\d\d'
+        \ ? expand('%:r')
+        \ : strftime('%F')
+  call wiki#url#parse('journal:' . l:date[:3] . '_m' . l:date[5:6]).open()
+endfunction
+
+" }}}1
+function! wiki#journal#make_index() " {{{1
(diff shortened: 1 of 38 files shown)


lervag:
Hi @kwrobert! Thank you for the PR!

First, I think you are addressing two issues with this. I would prefer if you could separate this, because I don't think your solution regarding month names is correct, although I agree it is an issue. If you open a new issue with a very short description of this, then I can implement a better solution that allows others (e.g. me) to keep a localised version of month names.

Regarding your question, I don't really think there is anything like python path joining in Vim. But I also think that things typically work as expected if you just always use forward slashes. I.e., you can write expressions like  `l:root . '/' . l:filename` or similar, and it should still work on Windows in most cases. IIRC.

I've also added a couple of comments to your code.

autoload/wiki/journal.vim
```diff
@@ -93,3 +129,5 @@ function! s:get_links_generic(rx, fmt) abort " {{{1
 endfunction
 
 " }}}1
+"
```
lervag: Please remove extra lines.

autoload/wiki/journal.vim
```diff
@@ -54,7 +54,43 @@ function! wiki#journal#go_to_month() abort " {{{1
...
+      put =''
+      for entry in entries
+        put ='- ['.entry.']('.entry.')'
+      endfor
+      put =''
+    endfor
+  endfor
+endfunction
+
+" }}}1
```
lervag: Please separate the public functions from the private functions with a single line here.

autoload/wiki/journal.vim
```diff
@@ -54,7 +54,43 @@ function! wiki#journal#go_to_month() abort " {{{1
 endfunction
 
 " }}}1
+function! wiki#journal#make_index() " {{{1
+  let l:regex_days = '\d\{4}-\d\d-\d\d'
+  let entries=s:get_links_generic(l:regex_days, '%Y-%m-%d')
```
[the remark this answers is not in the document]
lervag: Note, both comments need more changes in later code as well.

lervag:
Please see the comments. I should probably have used the review feature, but I think you find the comments if you click "File changes".

Note, I think the code generally looks very good, and the comments are a little bit pedantic. Still, I very much like to keep my code consistent, and I hope you can update as asked, then rebase and force push.

Let me know if you need any further assistance on this.

kwrobert:
No problem, consistency is important! 

I've added the requested modifications, quick question though, as I'm a bit new to contributing to open source. When you say "rebase and force push", do you mean squash these last two commits into a single commit then force push up to my remote repo to rewrite the history there?

And I agree my "solution" to the whole month issue was more just a reflection of my personal preference. I rolled that back to the version from your repo, I'll open up a separate issue so we can discuss a more general way to handle this.

autoload/wiki/journal.vim
```diff
@@ -53,6 +53,43 @@ function! wiki#journal#go_to_month() abort " {{{1
...
+        let year_dict[month] = [entry]
+      endif
+    else
+      let sorted_entries[year] = {month:[entry]}
+    endif
+  endfor
+
+  for year in reverse(sort(keys(sorted_entries)))
+    let l:month_dict = sorted_entries[year]
+    put ='# '.year
```
lervag: Add spaces before/after the `.` (makes it easier to distinguish from dict members).

autoload/wiki/journal.vim
```diff
@@ -53,6 +53,43 @@ function! wiki#journal#go_to_month() abort " {{{1
...
+    put =''
+    for [month, entries] in items(month_dict)
+      let l:mname = wiki#date#get_month_name(month)
+      let l:mname = toupper(mname[0]).mname[1:strlen(mname)]
```
lervag: Same as above.

autoload/wiki/journal.vim
```diff
@@ -53,6 +53,43 @@ function! wiki#journal#go_to_month() abort " {{{1
...
+      put ='## '.mname
```
lervag: Ditto.

lervag:
Yes, when I say "rebase and force push", I mean that I prefer to get logical changes in separate commits, and that such changes typically require that you do `git push -f` to force push the changes on your remote repo.

And thanks, I think this looks mostly good. I had a couple of minor comments that escaped me the first time.

kwrobert:
Alright, I squashed all my new commits together. This should contain everything, for some reason Github closed my pull request after I force pushed. Mind reopening?

Also, two questions:

1) Should I add a command to expose this function to the user?
2) Right now the inserted journal links are relative, meaning they don't work if you were to insert them on any wiki page that isn't in the journal subdirectory. Should I fix this to make the links absolute?

lervag:
Hmm, that's strange. I can't reopen this PR. Perhaps you could just open a new one and link to this for the description?

PR #5, opened 2018-04-30
Added wiki#journal#make_index function

kwrobert:
Reopening from [here](https://github.com/lervag/wiki/pull/4), see link for the description.

### DIFF
--- a/autoload/wiki.vim
+++ b/autoload/wiki.vim
@@ -162,6 +162,7 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer WikiGraphFindBacklinks call wiki#graph#find_backlinks()
   command! -buffer WikiGraphIn            call wiki#graph#to_current()
   command! -buffer WikiGraphOut           call wiki#graph#from_current()
+  command! -buffer WikiJournalIndex       call wiki#journal#make_index()
   command! -buffer WikiLinkNext           call wiki#nav#next_link()
   command! -buffer WikiLinkOpen           call wiki#link#open()
   command! -buffer WikiLinkOpenSplit      call wiki#link#open('vsplit')
@@ -172,7 +173,6 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer WikiPageDelete         call wiki#page#delete()
   command! -buffer WikiPageRename         call wiki#page#rename()
   command! -buffer WikiPageToc            call wiki#page#create_toc()
-  command! -buffer WikiJournalIndex       call wiki#journal#make_index()
 
   if b:wiki.in_journal
     command! -buffer -count=1 WikiJournalPrev       call wiki#journal#go(-<count>)
@@ -189,6 +189,7 @@ function! s:init_buffer_mappings() abort " {{{1
   nnoremap <buffer> <plug>(wiki-graph-find-backlinks) :WikiGraphFindBacklinks<cr>
   nnoremap <buffer> <plug>(wiki-graph-in)             :WikiGraphIn<cr>
   nnoremap <buffer> <plug>(wiki-graph-out)            :WikiGraphOut<cr>
+  nnoremap <buffer> <plug>(wiki-journal-index)        :WikiJournalIndex<cr>
   nnoremap <buffer> <plug>(wiki-link-next)            :WikiLinkNext<cr>
   nnoremap <buffer> <plug>(wiki-link-open)            :WikiLinkOpen<cr>
   nnoremap <buffer> <plug>(wiki-link-open-split)      :WikiLinkOpenSplit<cr>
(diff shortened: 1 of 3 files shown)


lervag:
Replies to your [questions](https://github.com/lervag/wiki/pull/4#issuecomment-385393883):

1. I don't find it strictly necessary, but I think it could make sense with respect to documentation. That is, if you add a command like `WikiJournalGenerateIndex` or similar and then document it in a similar fashion as the other commands. One should probably also add a corresponding `<plug>` mapping.

2. The links should be of the form `journal:<name>`, e.g. `journal:2014-03-02`. This will work regardless of where you are. Also, I noticed that you add `[text](url)` type links, but I think that you should use `[[url|text]]` instead. At least, that's what I've used as a standard for the wiki type links. I use `[..](..)` as the standard syntax for external links. [Note, though, that both types work equally well, so this is mostly a cosmetic matter.]

kwrobert:
Ah, the reason I use `[text](url)` style links is to be compatible with [Pandoc](https://pandoc.org/MANUAL.html#links) conversions, and also to allow automatic rendering of my notes into a beautiful Wikipedia style format viewable from my browser with [Gitit](https://github.com/jgm/gitit). 

I actually have some opinions/ideas about Pandoc and Vimtex compatibility, but perhaps this issue is not the appropriate place to discuss it. I would actually like to hear about your workflow for using this wiki you've created (if you're interested in sharing, that is). Do you have a method for rendering your notes?

autoload/wiki.vim
```diff
@@ -172,6 +172,7 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer WikiPageDelete         call wiki#page#delete()
   command! -buffer WikiPageRename         call wiki#page#rename()
   command! -buffer WikiPageToc            call wiki#page#create_toc()
+  command! -buffer WikiJournalIndex       call wiki#journal#make_index()
```
lervag: I've ordered the commands alphabetically. Not important, but I'd appreciate if you could follow the same idea.

autoload/wiki.vim
```diff
@@ -198,6 +199,7 @@ function! s:init_buffer_mappings() abort " {{{1
   nnoremap <buffer> <plug>(wiki-page-delete)          :WikiPageDelete<cr>
   nnoremap <buffer> <plug>(wiki-page-rename)          :WikiPageRename<cr>
   nnoremap <buffer> <plug>(wiki-page-toc)             :WikiPageToc<cr>
+  nnoremap <buffer> <plug>(wiki-journal-index)        :WikiJournalIndex<cr>
```
lervag: Similar as above.

autoload/wiki/journal.vim
```diff
@@ -53,6 +53,43 @@ function! wiki#journal#go_to_month() abort " {{{1
   call wiki#url#parse('journal:' . l:date[:3] . '_m' . l:date[5:6]).open()
 endfunction
 
+" }}}1
+function! wiki#journal#make_index() " {{{1
```
lervag: Could we add an argument to support different link styles? Something like `make_index(use_md_links)`?

autoload/wiki/journal.vim
```diff
@@ -53,6 +53,43 @@ function! wiki#journal#go_to_month() abort " {{{1
...
+    let l:month_dict = sorted_entries[year]
+    put ='# ' . year
+    put =''
+    for [month, entries] in items(month_dict)
+      let l:mname = wiki#date#get_month_name(month)
+      let l:mname = toupper(mname[0]) . mname[1:strlen(mname)]
+      put ='## ' . mname
+      put =''
+      for entry in entries
+        put ='- [' . entry . '](journal:' . entry . ')'
```
lervag: Following the above comment: Could we add two versions here:

```vim
if a:use_md_links
  put ='- [' . entry . '](journal:' . entry . ')'
else
  put ='- [[journal:' . entry . '|' . entry ']]'
endif
```

autoload/wiki.vim
```diff
@@ -162,6 +162,7 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer WikiGraphFindBacklinks call wiki#graph#find_backlinks()
   command! -buffer WikiGraphIn            call wiki#graph#to_current()
   command! -buffer WikiGraphOut           call wiki#graph#from_current()
+  command! -buffer WikiJournalIndex       call wiki#journal#make_index()
```
lervag: Ok, so, the last part is to add the argument to the command. Either two versions of the command, or perhaps better, a "bang" command, i.e. something like this:

```vim
command! -buffer -bang WikiJournalIndex call wiki#journal#make_index(<q-bang> == '!')
```

This also requires a minor expl. in the docs, as well as two versions of the `<plug>` mapping.

lervag: I.e., `WikiJournalIndex` uses wiki type links, and `WikiJournalIndex!` uses markdown links.

kwrobert: Ah! Right, sorry I missed this. One moment

kwrobert:
Does this look right? This whole 'bang' command thing was new to me. 

Also, thanks for walking me through this so patiently! This is actually my first pull request ever and it's been a great experience.

doc/wiki.txt
```diff
@@ -241,6 +241,16 @@ the commands are also available as mappings of the form `<plug>(wiki-[name])`.
 *WikiPageToc*
   Create/Update table of contents.
 
+*<plug>(wiki-journal-index)*
+*WikiJournalIndex*
+  Insert a sorted list of links to all journal pages below the cursor using
+  wiki style links
+
+*<plug>(wiki-journal-index)*
```
lervag: Sorry, but it seems you forgot `-md` here.

lervag:
It's basically right, but with a minor remaining comment. I'll merge when it's fixed.

And I'm happy to hear that you're not put off by my nitpicking. :)

lervag:
Thanks!

PR #7, opened 2018-05-16
Use g:wiki_journal_format to create new journal pages

### LINKED ISSUE #2: Weekly/monthly journal support
Similar to https://github.com/vimwiki/vimwiki/issues/265, I'd like to express my interest in a tweak to the journal functionality to allow it to use weekly journal files.

To provide some flexibility here, do you think it is feasible to allow specifying a custom date format string with which to create the journal file? For example, if the default is `%Y-%m-%d` which would be 2017-11-27 today and 2017-11-28 tomorrow, and you wanted to reduce the frequency of journal entries to weekly, you could specify `%Y-%U` which would be `2017-48` this week and `2017-49` next week, or `%Y-%m` for monthly for example.

Please let me know your thoughts!

jgilchrist:
I'm trying to address the (very small) number of issues that mean I currently maintain my own fork of this plugin. At the moment those are:

* Weekly journals
* 'Summary' as a title instead of 'Innhand'
* ~.wiki filetype detection does not always work (although I haven't tested this more recently)~ I tested this again and it works

I'm unsure of the approach here so please feel free to point me in a better direction if you'd prefer this solved in a different way.

Fixes #2

### DIFF
--- a/autoload/wiki/journal.vim
+++ b/autoload/wiki/journal.vim
@@ -6,7 +6,9 @@
 "
 
 function! wiki#journal#make_note(...) abort " {{{1
-  let l:date = (a:0 > 0 ? a:1 : strftime('%Y-%m-%d'))
+  let l:date = a:0 > 0
+      \ ? a:1
+      \ : strftime(get(g:, 'wiki_journal_format', '%Y-%m-%d'))
   call wiki#url#parse('journal:' . l:date).open()
 endfunction
 
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -117,6 +117,11 @@ CONFIGURATION                                                     *wiki-config*
 
   Default: `'journal'`
 
+*g:wiki_journal_format*
+  The format string used to construct new journal files.
+
+  Default: `'%Y-%m-%d'`
+
 *g:wiki_projects*
   List of project names (or regexes) used to collect journal entries when
   creating weekly or monthly summaries.


lervag:
Thanks!

I have some comments:

- This does address the journal format, it seems. I would be happy if you could explain a little bit more about the motivation here.
- This does not seem to address the "Summary" as a title instead of "Innhald"?
- You should add an entry to the docs under the `CONFIGURATION` section. Preferably after the `g:wiki_journal` entry.

autoload/wiki/journal.vim
```diff
@@ -6,7 +6,7 @@
 "
 
 function! wiki#journal#make_note(...) abort " {{{1
-  let l:date = (a:0 > 0 ? a:1 : strftime('%Y-%m-%d'))
+  let l:date = (a:0 > 0 ? a:1 : strftime(get(g:, 'wiki_journal_format', '%Y-%m-%d')))
```
lervag: I'd prefer to keep lines at 79 columns. My proposed formatting would be:

```vim
let l:date = a:0 > 0
    \ ? a:1
    \ : strftime(get(g:, 'wiki_journal_format', '%Y-%m-%d'))
```

jgilchrist:
Ah yes, sorry for the misunderstanding, this was only intended to address the first bullet point.

I covered some of my motivation in #2 but generally I find that daily journals are too frequent for me and it's more helpful to collate notes on a weekly basis. For example, I have meetings where it's useful to refer back to things that happened during the week and having them all available in a single document is more convenient. I've maintained my own fork for weekly journals and have generally found it's suited my workflow better than the default option (I've also seen similar requests on the vimwiki repository - https://github.com/vimwiki/vimwiki/issues/265).

I've addressed both your comments - please let me know if there's anything else you'd like me to change!

lervag:
Note: Stepping up/down might already work due to the way it is implemented. Perhaps your suggested change is all that is needed for the basic features to work as expected.

One thing that might be worth a thought: I personally write daily entries, and I collect weekly and monthly brief summaries. For this, I use the `template.vim` code, as well as the `journal.vim:go_to_[week|month]` functions. Should we update this accordingly?

For #2, I was thinking of implementing a resolution setting on day/week/month, and if you set it to week, the default would be that opening a day would go to that days week. Would such an update solve what you have solved here?

lervag:
> 'Summary' as a title instead of 'Innhand'

This is now fixed in #20.

Issue #20, opened 2019-03-12
Journal commands create Norwegian(?) date names


jounathaen:
When I create a journal monthly page by using `WikiJournalToMonth`, It adds the the caption in norwegian:

> Samandrag fra mars 2019

The same happens with the month names in the output of `WikiJournalIndex`

Apparently the date names are hardcoded in norwegian(?).

Additional Issue (maybe I have to create a new issue):
Using `WikiJournalToWeek` creates a buffer with the filename `2019_wdate: illegal time format.wiki`

lervag:
Oh, sorry about that. The lack of complaints have made me forget about fixing this. I'll make it a priority to update this now!

jounathaen:
I created a pull-request #22

lervag:
Ok, I've pushed a first version. I think the updated docs should be easily understandable. Let me know what you think.

lervag:
I'll look at the issue with the bad buffer filename. Could you give me the output when you do e.g. `echo wiki#date#get_week('2019-03-13')`? I suspect there might be something wrong here. If the output is strange, please next try `date +%V -d 2019-03-13` in a terminal. Also, it would be interesting to see the output of `date --version`.

jounathaen:
```
echo wiki#date#get_week('2019-03-13')
date: illegal time format
```
```
> date +%V -d 2019-03-13
date: illegal time format
usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```
`date --version` does not work, but in the manpage I see, that it is the BSD implementation (Macbook). On a Linux machine it works.

To get the weeknumber from a given date in BSD, apparently you have to use:
```
$ date -j -f '%Y-%m-%d' '2019-03-13' +%V
11
```
Which does not work on my Linux machine anymore... 
(apparently others already had simmilar problems https://stackoverflow.com/questions/8747845/how-can-i-detect-bsd-vs-gnu-version-of-date-in-shell-script)

lervag:
Sorry about the delay. I've tried to add the BSD syntax as a fallback. Let me know if it works.

jounathaen:
Seems to work 👍

jounathaen:
Hmm, unfortunately not.
[...]

PR #22, opened 2019-03-12
Modified the get_month_name function, so that it uses the name from the current locale

jounathaen:
If no python is found, or no locale is set, it defaults to en_US.

This is my first contribution in vimscript, so please review.
I'm not 100% happy with the duplicate python code.

Also I only tested it on neovim and not on vim.

### DIFF
--- a/autoload/wiki/date.vim
+++ b/autoload/wiki/date.vim
@@ -64,20 +64,46 @@ endfunction
 
 " }}}1
 function! wiki#date#get_month_name(month) abort " {{{1
-  return get([
-        \ 'januar',
-        \ 'februar',
-        \ 'mars',
+  " get the monthname from python which has a broad locale support
+  " pythonx seems not to be available on all vim setups, therefore the duplicate is
+  " necessary
+  if has('python')
+    python << EOF
+import vim, locale, calendar
+try:
+  locale.setlocale(locale.LC_ALL, '')
+except locale.Error: # " Fallback locale, if none is set
+  locale.setlocale(locale.LC_ALL, 'en_US')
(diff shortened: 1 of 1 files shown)


jounathaen:
Should fix #20

lervag:
I very much appreciate your effort here! But: I much prefer to use more robust solutions with vimscript (when possible). E.g., in this case, I will rather let english be the default and add an option to allow to customize the month names. This will also generally give more control to the user. I'll implement a solution now, then we can discuss it in #20 when I push it. If you should have some good reasons to make it more complex than what I propose than of course, I'm all ears.

jounathaen:
Well, this approach falls back to vimscript, if there is no python available.

Currently vim is available in 8 languages and manually adding the month names is surely more code than the 5 lines of python. Personally I'm fine with english, but if not, adding lines for the localization of each plugin would be annoying and a lot of lines in my vimrc. I don't think that localization should be an user option.

But that is of course just my opinion and I think your implementation is doing the job just fine. I just wanted to justify my approach ;-)

lervag:
Ok, a suggestion for improving the current implementation: What if the `g:wiki_month_names` variable is by default initialized through python similar to what you suggest, unless it is specified by the user? Also, if the python method fails, we fall back to the current default list.

jounathaen:
The longer I think about it, the more I don't care :-D
But as a suggestion for improvement, this could be kept in mind.

lervag:
Thanks. I'm personally happy with the way it works now, and I like to keep things simple. Thus I'll leave it as it is for now, until someone should bump this or open a related issue.

PR #30, opened 2019-04-03
Config to pick default link type

CrossR:
Hey!

Vimscript isn't my strongest language, so shout up if something is done wrong/inefficiently/ugly.

This just adds a config option to pick the default link style, which is currently just `wiki` or `md`.

TODO:

 - [x] : Get feedback on a good config option name.
 - [x] : Write docs (pending getting a name okay-ed)
 - [x] : Write a test?

I'll admit to not really knowing the workflow for tests in vimscript, so for now I've just done some manual ones.

One weird thing I saw was that on line 306 of `link.vim`, 3 arguments seem to be passed over, when I can only see the first and second being used?

### DIFF
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -242,8 +242,14 @@ function! wiki#link#template_md(url, ...) abort " {{{1
   return '[' . l:text . '](' . a:url . ')'
 endfunction
 
+" }}}1
 function! wiki#link#pick_template_type(url, ...) abort
-  if !empty(g:wiki_link_target_type) && exists('*wiki#link#template_' . g:wiki_link_target_type)
+  "
+  " Pick the relevant link template command to use based on the users
+  " settings. Default to the wiki style one if its not set.
+  "
+  if !empty(g:wiki_link_target_type) &&
+    \ exists('*wiki#link#template_' . g:wiki_link_target_type)
     return call('wiki#link#template_' . g:wiki_link_target_type, [a:url] + a:000)
   else
     return wiki#link#template_wiki(a:url, a:000)
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -242,6 +242,14 @@ function! wiki#link#template_md(url, ...) abort " {{{1
   return '[' . l:text . '](' . a:url . ')'
 endfunction
 
+function! wiki#link#pick_template_type(url, ...) abort
+  if !empty(g:wiki_link_target_type) && exists('*wiki#link#template_' . g:wiki_link_target_type)
+    return call('wiki#link#template_' . g:wiki_link_target_type, [a:url] + a:000)
+  else
+    return wiki#link#template_wiki(a:url, a:000)
+  endif
+endfunction
+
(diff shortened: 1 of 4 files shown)


lervag:
> One weird thing I saw was that on line 306 of link.vim, 3 arguments seem to be passed over, when I can only see the first and second being used?

You mean here, I guess:

```vim
  elseif len(l:candidates) == 1
    return wiki#link#template_wiki('/' . l:candidates[0], l:url, l:text)
  endif
```

This is a bug. I'm fixing.

autoload/wiki/link.vim
```diff
@@ -242,6 +242,20 @@ function! wiki#link#template_md(url, ...) abort " {{{1
...
+  " Pick the relevant link template command to use based on the users
+  " settings. Default to the wiki style one if its not set.
+  "
+  if !empty(g:wiki_link_target_type) &&
+    \ exists('*wiki#link#template_' . g:wiki_link_target_type)
+    return call('wiki#link#template_' . g:wiki_link_target_type, [a:url] + a:000)
+  else
+    return call ('wiki#link#template_wiki', [a:url] + a:000)
+  endif
+endfunction
```
lervag: I would use `template_pick_from_option` as the name and I would implement it more like this:

```vim
function! wiki#link#template_pick_from_option(...) abort
  let l:type = get(g:, 'wiki_link_target_type', 'wiki')

  try
    return call('wiki#link#template_' . l:type, a:000)
  catch /E117/
    echoerr 'Link target type does not exist:' l:type
    echoerr 'See ":help g:wiki_link_target_type" for help.'
  endtry
endfunction
```

plugin/wiki.vim
```diff
@@ -36,6 +36,7 @@ call wiki#init#option('wiki_index_name', 'index')
 call wiki#init#option('wiki_root', '')
 call wiki#init#option('wiki_link_extension', '')
 call wiki#init#option('wiki_link_target_map', '')
+call wiki#init#option('wiki_link_target_type', '')
```
lervag: Should initialize this with a value. Keep current default, so set it to `wiki`.

autoload/wiki/link.vim
lervag: Although: The option should be initialized when `wiki.vim` is initialized, so it seems to be not necessary to use the `get()` function.

lervag:
Ok, this seems quite finished now. Remaining question is if we should add a test for it. Of course, that would be optimal, but I understand it is not trivial.

I've tried to create a test system that works well, but testing vimscript is admittably a little bit tricky.

Let me know if you need some assistance for this.

CrossR:
Test system was easy once I worked out how to get the tests running!

I've added a test for setting the option and setting it with an extension, since I figure that is a pretty common usage. I wasn't 100% if that second test made more sense where I've put it, or over in the ` 	test_markdown.vim` file, but I figured it wouldn't matter too much.

lervag:
Great! This is very good, thanks. I really appreciate your effort!

lervag:
This has now been merged.

PR #35, opened 2019-08-10
Silently cancel completion if not link is found

andreypopp:
Before this commit completion func returned `-1` which means "cancel
completion with an error" vim and does nothing in neovim. This commit
makes it return `-3` which cancels the completion silently. That
unbreaks neovim which doesn't handle `-1` and proceeds with completion
fails on undefined `s:ctx` variable instead.

### DIFF
--- a/autoload/wiki/complete.vim
+++ b/autoload/wiki/complete.vim
@@ -9,7 +9,7 @@ function! wiki#complete#omnicomplete(findstart, base) " {{{1
   if a:findstart
     let l:line = getline('.')[:col('.')-2]
     let l:cnum = match(l:line, '\[\[\zs[^\\[\]]\{-}$')
-    if l:cnum < 0 | return -1 | endif
+    if l:cnum < 0 | return -3 | endif
 
     let l:base = l:line[l:cnum:]
 


lervag:
Great, thanks for noticing and fixing!

Btw, when I compare the docs for Vim and neovim, it seems they are the same:

```
# From Vims ":help complete-functions"
Negative return values:
   -2 	To cancel silently and stay in completion mode.
   -3 	To cancel silently and leave completion mode.
   Another negative value: completion starts at the cursor column
```

```
# From neovims ":help complete-functions"
Negative return values:
   -2 	To cancel silently and stay in completion mode.
   -3 	To cancel silently and leave completion mode.
   Another negative value: completion starts at the cursor column
```

I expect that Vim and neovim somehow have different behaviour for what happens when things are "buggy" with the completion function. In any case, this is a good catch!

andreypopp:
I compared against an old vim so they might have fixed it then.

Thank you!

lervag:
No, thank you! :) I'm happy for contributions to this plugin. I'm sorry I can't maintain it/further develop it at a more steady rate. But at least I think it does mostly do its job...

andreypopp:
>  I'm sorry I can't maintain it/further develop it at a more steady rate. But at least I think it does mostly do its job..

No worries — it's perfectly usable (at least for my needs) and I'm still learning its features! Like it very much so far.

lervag:
Thanks! Feel free to open issues if you have ideas for improvement or come over a bug!

Issue #38, opened 2019-11-07
instructions for running tests


edrex:
I spent 10 minutes poking at the makefile but I can't figure out how to run the tests. I would update doc if I knew, so if you post quick instruction here i'll clean up and PR

edrex:
Also, ty for your work on this. I used vimwiki for a week before I started contemplating reimplementing the core functionality without a ftplugin, so glad someone else did it first :)

lervag:
> I spent 10 minutes poking at the makefile but I can't figure out how to run the tests. I would update doc if I knew, so if you post quick instruction here i'll clean up and PR

Hah, sorry. I've been wanting to work more on this, but currently it is at a relatively immature state (testing, that is).

> Also, ty for your work on this. I used vimwiki for a week before I started contemplating reimplementing the core functionality without a ftplugin, so glad someone else did it first :)

Happy to hear it! I'm still using this very actively, and I appreciate any help with improving it. The current state of wiki.vim is quite good wrt. the features I personally need, but I think there are several possible improvements to be made that I would also benefit from.

----

So, to testing. For me, the following works:

```sh
cd /path/to/wiki.vim/test
make
```

The output is

```
> make
nvim -u test_journal.vim
nvim -u test_markdown.vim
nvim -u test_list.vim
nvim -u test_links.vim
nvim -u test_init.vim
```

Note, though, that it runs both `nvim` and `vim`. So perhaps the tests fail on your end because you don't have `nvim` installed?

The Makefile looks like this:

```make
export QUIT = 1

tests := $(wildcard test_*.vim)

.PHONY: $(tests)

all: $(tests)

$(tests):
	nvim -u $@
	@vim -n -u NONE -e -s -c "source $@" 2>&1
```

The idea is to run all `test_*.vim` files, e.g. `test_journal.vim`, with `nvim -u test_journal.vim`. Since the environment variable `QUIT` is set to 1, the function `wiki#test#quit()` ensures that the script exits when it is finished.

edrex:
I have nvim and a symlink vim -> nvim. If I comment the `@vim` line I get

```
Error detected while processing /home/eric/.vim/plugged/wiki.vim/test/init.vim:
line    6:
E117: Unknown function: wiki#test#init
Error detected while processing /home/eric/.vim/plugged/wiki.vim/test/test_init.vim:
line   12:
E117: Unknown function: wiki#test#error
```
 and then vim comes up and must be exited for each test file.

[...]

PR #39, opened 2019-11-07
md link template text: fall back to url if empty

edrex:
This resolves one nit I had when I started using wiki.vim:
The interactive title prompt on link toggle defaults to empty.

For me, probably others, 98% or so of the time I just want to use the link target
as the title, like vimwiki does.

This looks like what's expected in https://github.com/lervag/wiki.vim/blob/master/test/test_markdown.vim#L27 but I guess that's the old vimwiki code path.

Another idea would be to add a setting var to disable the prompt, but I'd always rather fix the default than add another esoteric setting that increases the spread of user configs etc.

ps idk how to run tests #38

### DIFF
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -240,6 +240,9 @@ function! wiki#link#template_md(url, ...) abort " {{{1
   let l:text = a:0 > 0 ? a:1 : ''
   if empty(l:text)
     let l:text = input('Link text: ')
+    if empty(l:text)
+      let l:text = a:url
+    endif
   endif
   return '[' . l:text . '](' . a:url . ')'
 endfunction


lervag:
> This resolves one nit I had when I started using wiki.vim:
> The interactive title prompt on link toggle defaults to empty.

> For me, probably others, 98% or so of the time I just want to use the link target
> as the title, like vimwiki does.

Do I understand correctly that you are using `wiki.vim` with `g:wiki_link_target_type` set to `md`?

Further, that you want `<cr>` to work like this:

1. Assume cursor is on top of a word `something`,

2. Press `<cr>`

3. `something` becomes `[something](something.md)`

For me, this seems to already work that way. As such, it would be useful if you could provide a minimal example that better explains what your initial problem/nit is.

----

Or, perhaps this is a way to reproduce:

0. Standard settings.
1. Click `<cr>` on `something` -> `[[something]]`.
2. Click `<leader>wf` to toggle to markdown type link, which now asks for input.

Instead of asking for input, you are suggesting that it instead fills the text with the link, so step 2 gives `[something](something)`.

Do I understand correctly?

edrex:
> Do I understand correctly that you are using wiki.vim with g:wiki_link_target_type set to md?

yes, i have 

```
let g:wiki_link_target_type = 'md'
let g:wiki_filetypes = ['md']
```

> For me, this seems to already work that way. As such, it would be useful if you could provide a minimal example that better explains what your initial problem/nit is.

for me, a "link text" prompt appears in status line. if I `<CR>` or `ESC` without inputting text, the output link text is blank.

Maybe there's another setting coming into play here? Those two and `g:wiki_root` are the only wiki vars I have set.

> Or, perhaps this is a way to reproduce:

>    Standard settings.
>    Click <cr> on something -> [[something]].
>    Click <leader>wf to toggle to markdown type link, which now asks for input.

I haven't tried with standard settings

> Instead of asking for input, you are suggesting that it instead fills the text with the link, so step 2 gives `[something](something)`.

I think that would be a reasonable behavior, for my usage habits, I rarely want to change the link text and when I do I can do it after creating the link, but idk if other ppl prefer the prompt. I wish we could instrument or survey users :D. Anyway because I didn't want to remove functionality people might like, this change just gives the prompt a default value. I would be happy with just removing the prompt.

edrex:
I tested with wiki and indeed there's no prompt, which makes sense because the wiki syntax doesn't require a separate link text. Propose changing md to work the same, as you sketched above.

edrex:
I've been playing with the tests, and would like to add some more. I would like to see better output for value assertions. @lervag would you accept a patch to port the tests to https://github.com/junegunn/vader.vim? I don't care about the BDD syntax so much (although it is nice) but the value assertion output makes working with the tests much nicer.

lervag:
> I tested with wiki and indeed there's no prompt, which makes sense because the wiki syntax doesn't require a separate link text. Propose changing md to work the same, as you sketched above.

Good, agreed.

> I've been playing with the tests, and would like to add some more. I would like to see better output for value assertions. @lervag would you accept a patch to port the tests to https://github.com/junegunn/vader.vim? I don't care about the BDD syntax so much (although it is nice) but the value assertion output makes working with the tests much nicer.

Actually, I've recently started to avoid Vader as a framework, because I've realized I find manual testing with Makefiles better. I've updated all the tests in vimtex to a Makefile based system, and I'm very happy for it. Reason is, it makes it much easier for me to keep track of things and to separate the tests. Make sure that the tests are fully independent, and so on.

But I would very much appreciate help with adding more tests. Could I then ask that you instead of depending on Vader added a function or two to `wiki#test`, e.g. like [this](https://github.com/lervag/vimtex/blob/master/autoload/vimtex/test.vim). Here I rely on the `cquit` command to quit with an error code.

edrex:
i peeked at your vimtex plugin and local assert helpers after i wrote that and guessed what your response was likely to be. Sounds good.

lervag:
I agree the old syntax for tests in wiki.vim was not very good, but I think that the updated tests are much better. Further, for me, it is a very strong benefit that I have _full_ control over the tests. With Vader, I sometimes couldn't figure out what was happening, and it was difficult to debug this through Vader. With the current tests, it is much easier to debug. Of course, this is just my humble opinion.

PR #40, opened 2019-11-07
make tests set runtimepath relative to sfile so it works without vundle

edrex:
lifted from https://github.com/junegunn/vader.vim/blob/master/test/vimrc

### DIFF
--- a/test/init.vim
+++ b/test/init.vim
@@ -1,5 +1,5 @@
 set nocompatible
-let &runtimepath = '~/.vim/bundle/wiki.vim,' . &runtimepath
+let &runtimepath = expand('<sfile>:p:h:h') . ',' . &runtimepath
 filetype plugin indent on
 syntax enable
 


edrex:
note this blows away the path rather than appending but that seems ok given that we are trying to make repeatable tests and probably don't want the system plugins.

test/init.vim
```diff
@@ -1,5 +1,5 @@
 set nocompatible
-let &runtimepath = '~/.vim/bundle/wiki.vim,' . &runtimepath
+let &runtimepath = expand('<sfile>:p:h:h') . ',' . &runtimepath
```
lervag: Even simpler:

```vim
let &runtimepath = '../,' . &runtimepath
```

I.e., no reason not to just use relative paths here, is it?

lervag:
Note: This had _nothing_ to do with Vundle or vim-plug, per se, only the fact that I've kept my plugins in `~/.vim/bundle`. I did that already when I started with pathogen many years ago. Now I use vim-plug and the same directory structure, but there is no difference.

Still, I agree that the hardcoded path is an anti pattern.

test/init.vim
edrex: sure, looks good

edrex:
Fixed in 0ac5325

PR #52, opened 2020-01-24
Update wiki.txt

Ginner:
On line 499(500) I've changed "... `pdf` files."  to  "... output files." to clarify that not only pdf are available (as hinted with 'test.html'). Here "... exported files." could also be an option. I don't know what would be best.

On line 687 I change `wiki_pdf_viewer` to `wiki_viwer` as this seems to be the correct one (not `wiki_file_open`, right?). `wiki_pdf_viewer` is not found  anywhere else in the doc.

Thanks for this awesome plugin!

### DIFF
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -198,7 +198,7 @@ CONFIGURATION                                                     *wiki-config*
   Default: >
 
     let g:wiki_journal = {
-        \ 'name': 'journal'
+        \ 'name': 'journal',
         \ 'frequency': 'daily',
         \ 'date_format': {
         \   'daily' : '%Y-%m-%d',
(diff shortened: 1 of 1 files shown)


lervag:
Great, thanks a lot! I really appreciate the help and the kind words.

Ginner:
No, thank you! I can't describe how much I appreciate these open source projects and the work people like yourself put into them (I'm a big fan of vimtex as well). It's a huge honor just to help what little I can :)

lervag:
That's very kind of you to say, and I appreciate it! I don't mind help, and if you want to attempt fixing or suggesting improvements, feel free to do so. I'll be happy to assist! Or, if you are not comfortable to attempt any vimscript, then feel free to keep things at suggestions/bug reports. :)

PR #53, opened 2020-01-24
Update wiki.txt

Ginner:
The `<c-j>` and `<c-k>` bindings doesn't seem to have been set. And as such doesn't appear in the list under "DEFAULT MAPPINGS". I've changed the description to include the wiki-commands instead.


lervag:
Yes, you're right. Thanks!

PR #66, opened 2020-04-13
Fix bug introduced in d9004ae

wbthomason:
d9004ae makes `WikiOpen` fail to open or create a page if no `wiki_link_target_map` is set.


lervag:
Thanks! (And sorry about the inconvenience!)

Issue #64, opened 2020-04-11
[Feature Request] Allow function to transform names of new pages


wbthomason:
# Description

It would be useful to have something like `g:wiki_link_target_map` for creating new pages - a variable that a user can set to a `FuncRef` which will be called to transform the names of new wiki pages created via `WikiOpen`/similar.

# Use case example

One example use (maybe the main use?) of this would be to append the date of creation to the filenames of new pages, transforming e.g. `my_new_page` to `my_new_page_20200411`, etc.

jabirali:
Some questions about how something like that would work:

- Considering your example, does this mean that instead of `[[my_page]]` links you always want to write `[[my_page_20200411]]` and so on? 
- If so, do you want calling `WikiLinkOpen` on a non-existing link like `[[my_page]]` to change the source to `[[my_page_20200411]]` to be consistent with the new page? 
- What about a to a non-existing target like `[[my_page_20200411]]` which occurs when a file has been deleted, that would in that case change the link to `[[my_page_20200411_20200412]]`?

wbthomason:
Yeah, those are good considerations. I had envisioned this feature as not explicitly interacting with links - between the existing `wiki_link_target_map` feature and the existing page omnicompletion, I think there are already good options both for users who prefer writing out the full `[[my_page_20200411]] ` and those who prefer writing `[[my_page]]` and having `WikiLinkOpen` resolve the correct target for them.

To my mind, adding this feature to page creation resolves an asymmetry between the two main interfaces to page paths: page creation and page linking.

lervag:
Would it suffice to ensure that `:WikiOpen` also uses `g:wiki_link_target_map`?

lervag:
I've pushed an update now. Please let me know if this was what you wanted.

wbthomason:
@lervag Thanks! I think this works for a sufficiently sophisticated `g:wiki_link_target_map` - for example, in the use case I give above, the function will need to check if the suffix is already present to avoid double-appends of the sort @jabirali mentions, the function will need to check if a page already exists (is being opened) or not (creating a new page), etc. 

Personally, I think it would still be cleaner to have separate variables for a `g:wiki_create_map` and `g:wiki_link_target_map`. I've poked around in the code a bit more and could probably put together a PR if you're open to this approach.

Incidentally, it looks like this change introduced a small bug - I submitted a fix in #66.

lervag:
[...]

PR #67, opened 2020-04-14
Add option for function to transform paths of new pages

wbthomason:
Per discussion in #64, this PR adds a very simple implementation of an option to specify a function which will be invoked to transform the paths of pages created through `WikiOpen`.

One design point that may need more iteration is that this runs the function for **all** invocations of `WikiOpen` - it does not test that the page is being created (rather than opened) first, and leaves this to the user (as shown in the example added to the docs). This may be good enough, or it may be better to add logic to ensure that the transformation function is only run for new pages.

### DIFF
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -316,8 +316,8 @@ function! wiki#link#template_word(url, ...) abort " {{{1
   "
   " Allow to map text -> url
   "
-  if !empty(g:wiki_link_target_map) && exists('*' . g:wiki_link_target_map)
-    let l:url = call(g:wiki_link_target_map, [a:url])
+  if !empty(g:wiki_map_link_target) && exists('*' . g:wiki_map_link_target)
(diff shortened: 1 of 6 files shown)


lervag:
Thanks, this looks good. I'll merge immediately, but I have one question: Would it be better if these options instead were named `g:wiki_map_link_target` and `g:wiki_map_create_page`? I'm thinking that it could make it more clear that these options are related with regard to their function, and as such, improve discoveribility?

wbthomason:
Yeah, that seems like a good idea. I'll push a commit with the renaming.

lervag:
You were fast, and I was already in the process of merging and making some minor modifications, so I made the change of option names myself.

Thanks, I appreciate your effort! It is very nice to get some extra hands on deck, as I've abit too much on my plate lately.

wbthomason:
Happy to help! Thanks!

PR #68, opened 2020-04-15
Add options to selectively enable only local/global default mappings

wbthomason:
This PR adds two configuration variables to allow users to choose a subset (local/global) of the default mappings to use. This is useful (to me, at least) because while the buffer-local mappings are nice to have, the global mappings interfere with existing keybindings in my config. I wasn't able to find a good way to enable only the local mappings (setting empty `lhs` values in `g:wiki_mappings_global` almost worked but caused strange `echo` output on loading the plugin); hence, this set of changes.

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -128,7 +128,7 @@ function! s:init_buffer_mappings() abort " {{{1
 
 
   let l:mappings = {}
-  if get(g:, 'wiki_mappings_use_all_defaults', 1) || get(g:, 'wiki_mappings_use_local_defaults', 0)
+  if index(['all', 'local'], g:wiki_mappings_use_defaults) >= 0
     let l:mappings = {
           \ '<plug>(wiki-code-run)' : '<leader>wc',
           \ '<plug>(wiki-graph-find-backlinks)' : '<leader>wb',
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -128,7 +128,7 @@ function! s:init_buffer_mappings() abort " {{{1
 
 
   let l:mappings = {}
-  if get(g:, 'wiki_mappings_use_defaults', 1)
+  if get(g:, 'wiki_mappings_use_all_defaults', 1) || get(g:, 'wiki_mappings_use_local_defaults', 0)
     let l:mappings = {
           \ '<plug>(wiki-code-run)' : '<leader>wc',
           \ '<plug>(wiki-graph-find-backlinks)' : '<leader>wb',
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -238,13 +238,31 @@ CONFIGURATION                                                     *wiki-config*
         \ },
         \}
 
-*g:wiki_mappings_use_defaults*
+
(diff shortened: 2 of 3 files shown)


autoload/wiki/buffer.vim
```diff
@@ -128,7 +128,7 @@ function! s:init_buffer_mappings() abort " {{{1
 
 
   let l:mappings = {}
-  if get(g:, 'wiki_mappings_use_defaults', 1)
+  if get(g:, 'wiki_mappings_use_all_defaults', 1) || get(g:, 'wiki_mappings_use_local_defaults', 0)
```
lervag, line 131: `get()` uses 0 as default value, so no need to use `, 0)` here.

lervag: Also, the line is too long. I want to keep things at 80 cols. So, change it to:

```vim
   if get(g:, 'wiki_mappings_use_all_defaults', 1)
        \ || get(g:, 'wiki_mappings_use_local_defaults')
```

doc/wiki.txt
```diff
@@ -238,13 +238,31 @@ CONFIGURATION                                                     *wiki-config*
...
+  this is set to 0, then only custom mappings will be applied. Custom mappings
   this is set to 0, then only custom mappings will be applied. Custom mappings
   may either be specified explicitly or through |g:wiki_mappings_global| and
+  may either be specified explicitly or through |g:wiki_mappings_global| and
   |g:wiki_mappings_local|.
+  |g:wiki_mappings_local|.
+
 
   Default: 1
+  Default: 1
```
lervag, line 254: The above seems unfinished. A lot of duplicate lines...

doc/wiki.txt
```diff
@@ -238,13 +238,31 @@ CONFIGURATION                                                     *wiki-config*
...
+
+
+*g:wiki_mappings_use_global_defaults*
+*g:wiki_mappings_use_local_defaults*
```
lervag, line 258: Perhaps we should collect the docs for all three variants? Also, perhaps it would be better to call it `g:wiki_mappings_use_defaults_{all,local,global}`?

lervag: Or, better: Instead of three options, how about a single option with different allowed values? It seems it could be enough with the values `'none', 'all', 'local', 'global'`, right?

doc/wiki.txt
wbthomason: Whoops - I have no idea how these lines got duplicated... Thanks for the catch.

doc/wiki.txt
wbthomason: Seems like a good plan, yep.

wbthomason: Fixed in a06e342

doc/wiki.txt
wbthomason: Fixed in a06e342

autoload/wiki/buffer.vim
wbthomason: Fixed in a06e342

wbthomason:
I implemented the faster version of your suggestion above in a06e342. Also fixed the strange duplication of lines in the docs.

lervag:
Thanks! I've merged this now.

PR #79, opened 2020-06-03
Add support for custom tag formats

wbthomason:
This PR adds a configuration variable `g:wiki_tag_format_pattern` to allow users to customize the format of their tags, e.g. if they would prefer a syntax like `#tag1 tag2` to the default `:tag1:tag2:`.

This PR is motivated by, when trying to use other editors for my Wiki files on platforms that do not support Vim (e.g. phones, etc.), running into a lack of support for VimWiki-style tags, though hashtag-tags are more broadly supported.

I've tested the example pattern to support that particular style of tag, and it seems to work correctly (e.g. it doesn't capture headers, link fragments, etc.), but I wouldn't be too surprised if I'd missed an edge case.

If you feel this is feature-creep/otherwise out of scope of the project, no problem.

### DIFF
--- a/autoload/wiki/tags.vim
+++ b/autoload/wiki/tags.vim
@@ -205,7 +205,7 @@ function! s:tags.gather_from_file(file) abort dict " {{{1
 
     while v:true
       let [l:tag, l:pos, l:col]
-            \ = matchstrpos(l:line, '\v%(^|\s):\zs[^: ]+\ze:', l:col)
+            \ = matchstrpos(l:line, g:wiki_tag_format_pattern, l:col)
       if l:col == -1 | break | endif
 
       call self.add(l:tag, a:file, l:lnum, l:pos)
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -353,6 +353,15 @@ CONFIGURATION                                                     *wiki-config*
         \ 'output' : 'loclist',
         }
 
+*g:wiki_tag_format_pattern*
+  A regular expression describing the search pattern for a single tag.
+
+  Default: >
+    let g:wiki_tag_format_pattern = '\v%(^|\s):\zs[^: ]+\ze:'
+
+  Example (to allow tags to be of the form "#tag1 #tag2"): >
+    let g:wiki_tag_format_pattern = '\v%(^|\s)#\zs[^# ]+\ze($| )'
+
 *g:wiki_template_title_month*
   A string that specifies the title of the month template. The following keys
   are interpolated:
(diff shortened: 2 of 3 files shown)


lervag:
> This PR adds a configuration variable `g:wiki_tag_format_pattern` to allow users to customize the format of their tags, e.g. if they would prefer a syntax like `#tag1 tag2` to the default `:tag1:tag2:`.

Great, this is very useful!

I took the liberty to change the name to `g:wiki_tags_format_pattern` to be consistent with `g:wiki_tags`. I prefer to keep the option names consistent, and if there were to be more options I think `g:wiki_tags` feels like a better prefix.

wbthomason:
Makes sense, thanks!

alwc:
@wbthomason What app are you using to browse the Markdown files on your phone? Does it support linking as well?

wbthomason:
@alwc I use Markor on Android and iA Writer on iOS. Markor supports links but does not automatically link tags; iA Writer supports links and automatically collects and links tags in the `#tag` format.

alwc:
Thank @wbthomason . I'll try out iA Writer on my iPhone

alwc:
Sorry for posting in this PR, but I got one more question for you @wbthomason 

I just installed iA Writer on my iPhone and it seems it doesn't support folder sync for Dropbox. I then tried to use iCloud and it seems the only way to sync my wiki.vim files is to move all of them to `~/Library/Mobile Documents/27N4MQEA55~pro~writer/Documents`. 

Is that how you setup your Markddown files for iOS? I suspect you use a different setup since you use both Android and iOS machines.

wbthomason:
@alwc: My setup is rather hacky; I can’t really recommend replicating it.

I primarily store my notes in Google Drive - that is, my `g:wiki_root` is my local Google Drive folder. This can be accessed and synced to from iA Writer (though I have run into one bug that caused Writer to remove a file from my Drive and move it to its own storage, this mostly has not seemed to happen). Markor for Android does not work with Google Drive, so I use Syncthing to *also* sync my wiki directory.

This is pretty bad practice and only works out alright because I’m only ever editing wiki files on a single device at a time, with ample time for things to sync in-between.

lervag:
@alwc @wbthomason I don't really view my wiki/notes on my phone. If I do, I do it with Vim on Android via Dropbox (I store my wiki on Dropbox). That works quite well, although it would be nice with a better integration. Still, not so important to me, as I tend to find a way to take note of things if I have to, then write it out in my wiki when I'm at my computer.

Issue #46, opened 2019-12-27
Is there any way to template new wiki pages?


Corey-Keller:
It's very possible that this is already a feature, and I just don't realize it, but is there any way to set up the default for new pages? For example, I turn the text "Example Page Here" into a link, and open it. 

Observed behavior: empty document

Expected behavior: A document starting with a level 1 heading containing the name of the link, and maybe a few new lines. Ideally, it would all be configurable beyond that, but that seems like a solid minimal default.

lervag:
Hi! Sorry for the delayed response; I've been mostly away from my computer during the Christmas break.

Thanks for raising the issue and giving a good description of what you want. This is not possible at the moment, but it should be a relatively simple feature to implement, and I'll look into it with the new year.

Do you have an idea of how you would want to configure such a feature? I could think of some alternatives, e.g.:

```vim
" String based template
let g:wiki_template_new = '# {pagename}'

" Function based template
let g:wiki_template_new = 'MyTemplate'

function! MyTemplate()
  return join([
    \ '# ' . expand('%:h'),
    \ '',
    \ 'Some text here.',
    \], "\n") . "\n"
endfunction
```

I think I prefer the last method, but it might be too complicated for a lot of people. If I implemented the function based feature, then I could also implement the string based version on top of it afterwards.

Corey-Keller:
I would think a file-based method would be best actually. the var would point to a file that would have (for example):
```
# {pagetitle}
last modified: {moddatetime}

IfSection[## {pagesection}

]
```

And it would read that and transform the variables appropriately:
(pagetitle} is auto-filled from the link description (if created that way and not blank)
{moddatetime} is the date and time the file was last modified (written to) in %Y-%m-%d %H:%M:%S (that's 2020-01-02 09:32:38 for example, but that format is just personal preference)
{pagesection} is in an "IfSection" block due to being optional, and would only be created for links with sections like "README.md#Section Two" for example. 

This is all hasty pseudo-code of course, but it lays out an idea of what I think would work best for sure. Gives the control of the function method, with _almost_ no changes from normal markdown.

lervag:
Ok, we could use a file based template. But to avoid a lot of future issues, I would want to make a good specification first. Here is a first draft:

**Option**  
We need a single option: `g:wiki_page_templates = [list of paths]`, where the paths are either relative to the current wiki page or to the root, `/`. That is, an absolute path will always be rooted at the wiki root.

**Template format**  
[...]

PR #83, opened 2020-06-21
Add template feature for new pages

lervag:
Tasks:

- [x] Write specification in the docs.
- [x] Implement feature according to the specification.
- [x] Write tests.

See also #46 for more context and discussion.

Feedback wanted: @Corey-Keller, @gauteh, @jabirali

Note: I think the specs are quite good now. I think the most important thing now is to agree on formats and structure, then we can add more interpolation features later, if we deem it necessary.

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -25,7 +25,7 @@ function! wiki#buffer#init() abort " {{{1
   call s:init_buffer_commands()
   call s:init_buffer_mappings()
 
-  call s:apply_template()
+  call wiki#template#init()
 
   if exists('#User#WikiBufferInitialized')
     doautocmd <nomodeline> User WikiBufferInitialized
@@ -170,22 +170,6 @@ function! s:init_buffer_mappings() abort " {{{1
   call extend(l:mappings, get(g:, 'wiki_mappings_local', {}))
 
   call wiki#init#apply_mappings_from_dict(l:mappings, '<buffer>')
-endfunction
-
-" }}}1
-
-function! s:apply_template() abort " {{{1
-  if filereadable(expand('%')) | return | endif
-
-  let l:match = matchlist(expand('%:t:r'), '^\(\d\d\d\d\)_\(\w\)\(\d\d\)$')
-  if empty(l:match) | return | endif
-  let [l:year, l:type, l:number] = l:match[1:3]
-
-  if l:type ==# 'w'
-    call wiki#template#weekly_summary(l:year, l:number)
-  elseif l:type ==# 'm'
-    call wiki#template#monthly_summary(l:year, l:number)
-  endif
 endfunction
 
 " }}}1
--- a/autoload/wiki/fzf.vim
+++ b/autoload/wiki/fzf.vim
@@ -98,6 +98,7 @@ function! s:accept_page(lines) abort "{{{1
 
   if len(a:lines) == 2 || !empty(a:lines[1])
     call wiki#page#open(a:lines[0])
+    sleep 1
   else
     let l:file = split(a:lines[2], '#####')[0]
     execute 'edit ' . l:file
--- a/autoload/wiki/nav.vim
+++ b/autoload/wiki/nav.vim
@@ -19,6 +19,22 @@ endfunction
 
 " }}}1
 
+function! wiki#nav#add_to_stack(link) abort " {{{1
+  let s:position_stack += [a:link]
+endfunction
+
+let s:position_stack = []
+
+" }}}1
+function! wiki#nav#get_previous() abort "{{{1
+  let l:previous = get(s:position_stack, -1, [])
+  if !empty(l:previous) | return l:previous | endif
+
+  let l:file = expand('#:p')
+  if filereadable(l:file) | return [l:file, 1] | endif
+endfunction
+
+" }}}1
 function! wiki#nav#return() abort "{{{1
   if g:wiki_write_on_nav | update | endif
 
@@ -30,12 +46,5 @@ function! wiki#nav#return() abort "{{{1
     silent! execute "normal! \<c-o>"
   endif
 endfunction
-
-" }}}1
-function! wiki#nav#add_to_stack(link) abort " {{{1
-  let s:position_stack += [a:link]
-endfunction
-
-let s:position_stack = []
 
 " }}}1
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -9,15 +9,7 @@ function! wiki#page#open(page) abort "{{{1
         \ !empty(g:wiki_map_create_page) && exists('*' . g:wiki_map_create_page)
         \ ? call(g:wiki_map_create_page, [a:page])
         \ : a:page
-  let l:url = wiki#url#parse('wiki:/' . l:page)
-
-  if !filereadable(l:url.path)
-    redraw!
-    call wiki#log#info('Opening new page "' . l:page . '"')
-    sleep 1
-  end
-
-  call l:url.follow()
+  call wiki#url#parse('wiki:/' . l:page).follow()
 endfunction
 
 "}}}1
--- a/autoload/wiki/template.vim
+++ b/autoload/wiki/template.vim
@@ -7,6 +7,21 @@
 function! wiki#template#init() abort " {{{1
   if filereadable(expand('%')) | return | endif
 
+  let l:context = {
+        \ 'date': strftime("%F"),
+        \ 'name': expand('%:t:r'),
+        \ 'origin': wiki#nav#get_previous(),
+        \ 'path': expand('%:p'),
+        \ 'path_wiki': wiki#paths#shorten_relative(expand('%:p')),
+        \ 'time': strftime("%H:%M"),
+        \}
+
+  for l:template in g:wiki_templates
+    if s:template_match(l:template, l:context)
+      return s:template_apply(l:template, l:context)
+    endif
+  endfor
+
   let l:match = matchlist(expand('%:t:r'), '^\(\d\d\d\d\)_\(\w\)\(\d\d\)$')
   if empty(l:match) | return | endif
   let [l:year, l:type, l:number] = l:match[1:3]
@@ -16,6 +31,28 @@ function! wiki#template#init() abort " {{{1
   elseif l:type ==# 'm'
     call wiki#template#monthly_summary(l:year, l:number)
   endif
+endfunction
+
+" }}}1
+
+function! s:template_match(t, ctx) abort " {{{1
+  if has_key(a:t, 'match_re')
+    return a:ctx.name =~# a:t.match_re
+  elseif has_key(a:t, 'match_func')
+    return a:t.match_func(a:ctx)
+  endif
+endfunction
+
+" }}}1
+function! s:template_apply(t, ctx) abort " {{{1
+  if has_key(a:t, 'source_func')
+    return a:t.source_func(a:ctx)
+  endif
+
+  let l:source = get(a:t, 'source_filename', '')
+  if !filereadable(l:source) | return | endif
+
+  call append(0, readfile(l:source))
 endfunction
 
 " }}}1
--- a/autoload/wiki/template.vim
+++ b/autoload/wiki/template.vim
@@ -35,6 +35,13 @@ endfunction
 
 " }}}1
 
+function! wiki#template#case_title(text, ...) abort " {{{1
+  return join(map(split(a:text), {_, x -> toupper(x[0]) . strpart(x, 1)}))
+endfunction
+
+" }}}1
+
+
 function! s:template_match(t, ctx) abort " {{{1
   if has_key(a:t, 'match_re')
     return a:ctx.name =~# a:t.match_re
@@ -52,7 +59,33 @@ function! s:template_apply(t, ctx) abort " {{{1
   let l:source = get(a:t, 'source_filename', '')
   if !filereadable(l:source) | return | endif
 
-  call append(0, readfile(l:source))
+  " Interpolate the context "variables"
+  let l:lines = join(readfile(l:source), "\n")
+  for [l:key, l:value] in items(a:ctx)
+    let l:lines = substitute(l:lines, '{' . l:key . '}', l:value, 'g')
+  endfor
+
+  " Interpolate user functions
+  let [l:match, l:c1, l:c2] = matchstrpos(l:lines, '{{[a-zA-Z#_]\+\s\+[^}]*}}')
+  while !empty(l:match)
+    let l:parts = matchlist(l:match, '{{\([a-zA-Z#_]\+\)\s\+\([^}]*\)}}')
+    let l:func = l:parts[1]
+    let l:arg = l:parts[2]
+    try
+      let l:value = call(l:func, [l:arg])
+    catch /E117:/
+      let l:value = ''
+    endtry
+
+    let l:pre = l:lines[:l:c1-1]
+    let l:post = l:lines[l:c2:]
+    let l:lines = l:pre . l:value . l:post
+
+    let [l:match, l:c1, l:c2] = matchstrpos(
+          \ l:lines, '{{[a-zA-Z#_]\+\s\+[^}]*}}', l:c2+1)
+  endwhile
+
+  call append(0, split(l:lines, "\n"))
 endfunction
 
 " }}}1
--- a/autoload/wiki/template.vim
+++ b/autoload/wiki/template.vim
@@ -4,6 +4,22 @@
 " Email:      karl.yngve@gmail.com
 "
 
+function! wiki#template#init() abort " {{{1
+  if filereadable(expand('%')) | return | endif
+
+  let l:match = matchlist(expand('%:t:r'), '^\(\d\d\d\d\)_\(\w\)\(\d\d\)$')
+  if empty(l:match) | return | endif
+  let [l:year, l:type, l:number] = l:match[1:3]
+
+  if l:type ==# 'w'
+    call wiki#template#weekly_summary(l:year, l:number)
+  elseif l:type ==# 'm'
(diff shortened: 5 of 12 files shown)


C-Fergus:
I see this hasn't really had much progress over the past year but for what it's worth, the specification describes most of what I would use it for.

I use two different styles of notes depending on if they're named "index" or not. This could be implemented by checking for the name of the file against a list of templates of the same filename, and failing that it could then fallback to a generic template. I understand that this may be a niche usecase though so I won't be distraught if it isn't implemented, this specification is good.

Apologies for not offering to help, I have made attempts to figure this out already but I don't know vimscript well enough.

lervag:
> I see this hasn't really had much progress over the past year but for what it's worth, the specification describes most of what I would use it for.

Thanks! That's worth a bunch. The lack of progress is mostly due to lack of interest. If I had a strong personal interest it would of course already have been implemented. But with external interest and input I would not mind picking up the threads and finishing this.

> Apologies for not offering to help, I have made attempts to figure this out already but I don't know vimscript well enough.

No problem! From my perspective, having a discussion to settle the discussion is in itself an important contribution!

> I use two different styles of notes depending on if they're named "index" or not. ...

You raise an interesting idea - that it should be possible to apply different templates depending on e.g. the target file name. How about the following refined idea?

```
let g:wiki_templates = [ { template_dict }... ]
```

That is, we define a list of templates whose "behaviour" is specified with a dictionary. The template feature would be implemented similar to this:

```vim
" new file is created first, then:
for template in g:wiki_templates
  if applies(template, context)
    call apply(template, context)
    break
  endif
endfor
```

Here the `applies` and `apply` functions would be implemented in wiki.vim and would rely on the attributes of the `template_dict`. E.g.:

* `template_dict.match_re`: applies is true if filename matches regex.
* `template_dict.match_function`: applies if `match_function` returns true with the `context` as argument (this is only applicable if `match_re` is not defined).
* `template_dict.source_file`: use a template file as the template source (as specified in the current specification).
* `template_dict.source_function`: use a custom function as the template source with the `context` as argument.

What do you think?

C-Fergus:
This seems like a better method, both for identifying the template to be used and for the templates themselves, as `source_function` has the potential to be more dynamic than a `.md` or `.rmd` file. `match_function` could also be used to solve the issue of identifying the containing directory in the following suggestion:

A suggestion based on the [discussion of the `pagetitle` definition](https://github.com/lervag/wiki.vim/issues/46#issuecomment-583891863) in #46, where the name of the file may not be the desired title for the document; can there be an additional context element and variable called `{directory}` or `{containing_folder}`, and the equivalent for the origin called `{last_directory}` or `{last_containingfolder}`? This would allow different templates based on the directory of the wiki containing the file, as well as allow the directory name to be used as the title as you suggested in that post.

An alternative might be to use the display text of the link used to create the file (i.e. `[THIS TEXT](...)` for markdown and `[[...|THIS TEXT]]` for wiki links). I don't know if this approach is any better, but it has the disadvantage(?) of being defined by the display text of the first link used to create the file, which isn't nearly as strict as using the name of the directory.

>No problem! From my perspective, having a discussion to settle the discussion is in itself an important contribution!

Great! Happy to help. Once I get around to sorting the HTML export and citations, these templates will be the final addition before I'll be content with my notes setup so I'm looking forward to it!

wasowski:
Just an irritating comment (so late in the process): I also initially found lack of templates a missing feature, but completely forgot about it half a year down the road.   It is very easy to create templates with snippets, so wiki.vim does not strictly need templates.

C-Fergus:
> ... It is very easy to create templates with snippets, so wiki.vim does not strictly need templates.

This was discussed in https://github.com/lervag/wiki.vim/issues/46#issuecomment-571803457, the advantage of templates is that they are automatic and for a lot of users would be much easier to set up if there is a requirement to access document properties as there is in the current spec.

> A suggestion based on the discussion of the pagetitle definition in #46, ...

Apologise for adding this after the initial suggestion too, I realise it also may have been quite irritating. Got a bit carried away with your enthusiastic response about having discussions!

anshul-march:
hi folks

so I wrote a Lua function to add date to my journal entries, I mentioned that here in #160 

as Karl had suggested that we should try to be compatible with Vim and older nVim, here's the same in VimL

```vim
fun! CurrentEntry()
  let filepath = g:wiki_root.'journal/'.strftime('%Y-%m-%d').'.md'
  if !filereadable(filepath)
    call writefile([strftime('# %a, %d %B %y')], filepath)
  endif
  exe 'edit' filepath
endfun
```

So while we're at it, can we have dynamic templates? What are you thoughts on this?

lervag:
> @C-Fergus This seems like a better method, ...

Good to hear we agree!

> A suggestion based on ...; can there be an additional context element and variable ...?

Yes, I think we can have several context elements. Let me start the implementation now with the current specification, then it should be easy to expand after.

> An alternative for the title might be to use the display text ... (i.e. `[THIS TEXT](...)` ... and `[[...|THIS TEXT]]` ...).

I propose to let this be "future work" for now; when we have a working version, then feel free to open a new issue with a feature request.

> > template_dict.match_function: applies if match_function returns true with the context as argument (this is only applicable if match_re is not defined).
> 
> I am not sure what you mean by "this is only applicable if match_re is not defined." In the example code ...

The point is that a specific "template" defines either `match_re` _or_ `match_func`, not both. If you happen to define both, then only `match_re` is used. The point of this is that `match_func` is more generic and can do exactly what `match_re` does, and so there is no reason to allow both at the same time.

In the example, the `g:wiki_templates` includes _three_ templates, the first two uses `match_re`, whereas the last one uses `match_func`.

> @wasowski Just an irritating comment (so late in the process): I also initially found lack of templates a missing feature, but completely forgot about it half a year down the road. It is very easy to create templates with snippets, so wiki.vim does not strictly need templates.

Agreed, but I also agree with @C-Fergus that the idea has its merits of being
automatic.

@anshul-march Thanks for the vimscript variant! Reg.

> So while we're at it, can we have dynamic templates? What are you thoughts on this?

I'm not fully sure what you mean. Can you be more specific?

anshul-march:
> > So while we're at it, can we have dynamic templates? What are you thoughts on this?
> 
> I'm not fully sure what you mean. Can you be more specific?

Let's take the example from the docs

```wiki
# {{wiki#template#case_title {pagetitle}}}
Created: {date} {time}

# Introduction

# Conclusion
```

I was thinking that just like `{date}`, we let user defined function to run inside the templates to generate text. This way users can customize the template however they want.

lervag:
> maybe there's already a way to do that, and I am just blind to it. Maybe the hint lies in wiki#template#case_title

Exactly: The proposed syntax is that `{{func arg}}` is interpolated to the value of `func(arg)`.

lervag:
Ok, it's starting to take shape. Still too early for real testing, e.g. no interpolation of the source files yet. But the source function should already work quite well. The code should be quite easy to read - see [here](https://github.com/lervag/wiki.vim/pull/83/files#diff-bc43bd5eaf654024e77004f97fbe04d15979bb22f91d0bde4ce7fa2475086fa7R7-R23).

lervag:
I believe the only thing left is to implement the template file functionality, i.e. variable and function interpolation, as well as the file search with the `;` suffix (as in `:help file-searching`). I will appreciate feedback already, but you should of course feel free to wait with testing until I've pushed a full implementation.

Please note the minor update of the specification; I've combined the template file variables with the context object.

C-Fergus:
> @lervag Please note the minor update of the specification; I've combined the template file variables with the context object.

Does this mean that the only difference between contexts and variables is function substitution? Is this difference arbitrary or can this also be added as a context to bring both to parity? Now that you've made this spec change I can't think of a reason to seperate variables and contexts other than implementation difficulties, but I would assume the easiest way to implement variables would be to either use the contexts defined in `wiki#template#init` or to call the same functions called by those contexts.

Apologies for any ambiguity in referring to "contexts," I mean to say the elements of the context dictionary.

lervag:
> Does this mean ...

I don't really understand what you're asking. What it means, is that the variables available in the text file format are exactly the keys of the context object. I.e., the table presented under `:help wiki-templates-context` also fully describes the available variables for the text file format described in `wiki-templates-format`. Does this answer your question?

lervag:
Ok, now I believe everything should work. It would be nice if someone wanted to test. I've simplified the template source path for now; if things work and you think it will be useful, then I will be glad to consider more flexible ways of specifying the template source file.

As the new features should not break any old features, and since it does not interfere, I believe it is safe to merge already and then apply changes on top afterwards. This also makes it easier for everyone to test, since you don't need to checkout the feature branch.

lervag:
I've pushed a couple of minor fixes/adjustments to the master branch as well, now.

Feel free to 1) continue discussing and giving feedback here, or 2) open new issues on specific problems or suggestions for the new template feature.

C-Fergus:
> What it means...

Okay I see from your last commit how the function substitution works. It answers my question. I copied the docs example code exactly to test but couldn't get it to work, I'll test again tonight.

anshul-march:
Hi Karl,

I tried using the template feature.

This is how I am setting templates for my daily journal and this works.

```vim
let g:wiki_templates = [
      \ { 'match_re': strftime('%Y-%m-%d'),
      \   'source_filename': '/Users/anshul/wiki/journal/.template.md'},
      \]
```

I do not understand quite clearly how to use `match_func`.

How do I pass filename to my function here.
```vim
fun! NewFunc(x)
  if a:x == 'rust'
    return v:true
  else
    return v:false
  endif
endfun

let g:wiki_templates = [
      \ { 'match_func': NewFunc(x),
      \   'source_filename': '/Users/anshul/wiki/.template-rust.md'},
      \]
```

Desired output: my `template-rust` to be sourced only when the name of the file is `rust`.

C-Fergus:
@anshul-march 

1. What I gather from `:help *g:wiki_templates*` ([see help page](https://github.com/lervag/wiki.vim/blob/master/doc/wiki.txt#L541)):

```vim
fun! NewFunc(context)
  if a:context.name == 'rust'
    return v:true
  else
    return v:false
  endif
endfun

let g:wiki_templates = [
      \ { 'match_func': function('NewFunc'),
      \   'source_filename': '/Users/anshul/wiki/.template-rust.md'},
      \]
```

However I haven't managed to get this working so I'm not sure if it is correct.

2. Use the `'path'` context key, from `:h *wiki-templates-context*` ([see help page](https://github.com/lervag/wiki.vim/blob/master/doc/wiki.txt#L1225)).

anshul-march:
Hi @C-Fergus I just tested your suggestions, and this worked. Thank you!

lervag:
@anshul-march

> This is how I am setting templates for my daily journal and this works. ...

I would adjust this, because you are not really defining a proper regex. This should be better, I think, as it should match _any_ file whose name matches an ISO date (but not a page that _contains_ a date):

```viml
let g:wiki_templates = [
      \ { 'match_re': '^\d\d\d\d-\d\d-\d\d$',
      \   'source_filename': '/Users/anshul/wiki/journal/.template.md'},
      \]
```

> I do not understand quite clearly how to use `match_func`.

I believe @C-Fergus answered this correctly. The point here is that the argument is a dictionary, and you need to access the correct keys (e.g. `name` and so on).

The value for the `'match_func'` must be a so called FuncRef, i.e. `function('FunctionName')`; as was also correctly fixed by @C-Fergus.

Note that the conditional can be trivially simplified:

```
if a:x == 'rust'
  return v:true
else                      ===>    return a:x == 'rust'
  return v:false
endif
```

Also, perhaps you wanted something "looser" than the name being exactly `rust`? E.g.:

```viml
function! NewFunc(ctx)
  return a:ctx.name =~ '[Rr]ust'
endfun
```

@C-Fergus

> However I haven't managed to get this working so I'm not sure if it is correct.

Feel free to ask if there is something in particular that does not work. It works for me, and it seems to work for @anshul-march..?

anshul-march:
> and it seems to work for @anshul-march..?

Yes, it works.

```vim
function! NewFunc(ctx)
  return a:ctx.name =~ '[Rr]ust'
endfun
```
this is great, I didn't even know one can do this

> as it should match any file whose name matches an ISO date 

yes, I didn't consider this case at all, thank you

I noticed that if i try to make a link with `[[Rust]]` it creates a file `Rust.md`.
And if try to make one with `[[rust]]` it creates `rust.md`

for uniformity, is there a way that the files are always created with lowercase?

what I mean, if I use `[[Rust]]` I get `rust.md`

lervag:
Yes; you want the `g:wiki_map_create_page` option. E.g.:

```vim
let g:wiki_map_create_page = 'ToLower'

function ToLower(name) abort
  return tolower(a:name)
endfunction

" In fact, since 'tolower()' already exists, you can probably use it directly
" without defining a custom function (but with a custom function you can do
" more, e.g. convert spaces to underscores):
let g:wiki_map_create_page = 'tolower'
```

lervag:
Sorry; this was partly wrong. You probably want that function, but you also want `g:wiki_map_link_create`.

lervag:
But; there is currently no feature to fully convert the link url. To do that, you need more heavy customization. So, the proper answer to your question is sort of no; sort of yes. To convert the wiki scheme url before opening it; i.e. make both `[[RUst]]` and `[[RUst|Text here]]` link to `rust.md`, then you need to change the `g:wiki_resolver` function. The default is here:

https://github.com/lervag/wiki.vim/blob/00502edf3052037ffbef59b88bace5d94f39ddf8/autoload/wiki/url/wiki.vim#L30-L61

So, this is more complex, but it is doable if you know some more Vimscript.

lervag:
Essentially, I guess one could implement a "converter" that is applied here as well similar to the `g:wiki_map_...` functions that transform the input `fname` or `url` or `link` (call it what you want) before passing it to the resolver function. I'm not sure if anyone would want to use it, though. I believe the `g:wiki_map_link_create` may be what you want, because it will enforce the lower case style and also allow you to keep the original text as the text part of the link.

C-Fergus:
> Feel free to ask if there is something in particular that does not work. It works for me, and it seems to work for @anshul-march..?

Sure, it's the regex that isn't working for me, everything else seems great. Using `match_func` works fine:

```vim
function! MatchIndex(dict)
  return a:dict.name == 'index'
endfunction

let g:wiki_templates = [
      \ { 'match_func': function('MatchIndex'),
      \   'source_filename': $HOME .  '/Documents/notes/templates/templateindex.rmd'},
      \]
```

Using `match_re` approximately as described in the docs does not work (the file is named "index.rmd" instead of "index.md"):

```vim
let g:wiki_templates = [
      \ { 'match_re': `index\.rmd'),
      \   'source_filename': $HOME .  '/Documents/notes/templates/templateindex.rmd'},
      \]
```

Just using 'index' as the regex match works but this also matches files that just have 'index' in the name which I don't want.

lervag:
Ah, yes - the regex doesn't work because the `name` attribute, which is the one matched against with `match_re`, does not contain the extension. So, in this particular case, you need to use the `match_func`. I think this is documented, but perhaps not clear enough? I pushed a minor clarification now.

C-Fergus:
> Ah, yes - the regex doesn't work because the `name` attribute, which is the one matched against with match_re, does not contain the extension. So, in this particular case, you need to use the `match_func`...

Ah okay I think I'm sorted now then, I created #167 to address the example code which prompted me to use the extensions.

Traap:
I've been unsuccessful using wiki templates.  I am certain the error is mine.  I have included my attempt to setup two templates.   I've read the doc and this thread multiple times.  Any help is appreciated.

```
let s:journal = g:wiki_root . '/.journal.md'
let s:template = g:wiki_root . '/.template.md'

let g:wiki_templates = [
    \ { 'match_re': '^\d\d\d\d-\d\d-\d\d$',
    \   'source_filename': s:journal},
    \ { 'match_re': '*',
    \   'source_filename': s:template}
    \]
```

I echoed g:wiki_templates does from both vim / neovim and the contents seams correct to me.  I have verity two templates are in my g:wiki_root folder too.  I get a blank .md file each time I create a new page.  :(

lervag:
There's a mistake in your `'match_re'`. Can you try this?

```vim
let s:journal = g:wiki_root . '/.journal.md'
let s:template = g:wiki_root . '/.template.md'

let g:wiki_templates = [
      \ { 'match_re': '^\d\d\d\d-\d\d-\d\d$',
      \   'source_filename': s:journal},
      \ { 'match_re': '.*',
      \   'source_filename': s:template}
      \]

" Or perhaps better:
let g:wiki_templates = [
      \ { 'match_re': '^\d\d\d\d-\d\d-\d\d$',
      \   'source_filename': s:journal},
      \ { 'match_func': {x -> v:true},
      \   'source_filename': s:template}
      \]
```

By the way, it would be useful to know your value of `g:wiki_root` as well.

Traap:
The first example worked for 2nd template only.   I used:

```
put=g:wiki_root
/home/traap/git/wiki

put=g:wiki_templates
g:wiki_templates:
{'match_re': '^\d\d\d\d-\d\d-\d\d$', 'source_filename': '/home/traap/git/wiki/.journal.md'}
{'match_re': '.*', 'source_filename': '/home/traap/git/wiki/.template.md'}
```

The second example did not work for either template:

```
put=g:wiki_root
/home/traap/git/wiki

put=g:wiki_templates
g:wiki_templates:
{'match_re': '^\d\d\d\d-\d\d-\d\d$', 'source_filename': '/home/traap/git/wiki/.journal.md'}
{'match_re': '*', 'source_filename': '/home/traap/git/wiki/.template.md'}
```

lervag:
I'm sorry, but this is a little bit hard to understand. Perhaps you can open a new issue to make the thread more focussed and easier to follow? I also think it's good to avoid "spamming" this feature thread with a possible configuration assist or bug report.

If/When you open a new issue, please note:

* I don't understand what you mean with `put=` to gather g:wiki values. It is much better to simply show me what you add to your vimrc file.

> The second example did not work for either template:

If I understand your `put` output correctly, it seems you did not really try my second alternative (`match_func` instead of `match_re`).

Issue #85, opened 2020-06-27
Request: scan tags at the middle/bottom of documents


ompugao:
Thank you for your invention and contribution to this great plugin!

wiki.vim requires document to contain tags at the first part of documents:
<https://github.com/lervag/wiki.vim/blob/a531d999d53f043f3c11bbc7ed8aa79bc9a31b7b/autoload/wiki/tags.vim#L193>

in my case, I always put tags at the middle/bottom, which wiki.vim does not recognize.

I can easily guess that the current implementation comes from your consideration on the speed of scanning.
Could you think about an option to change this behavior?

lervag:
> Thank you for your invention and contribution to this great plugin!

Happy you like it! :)

> wiki.vim requires document to contain tags at the first part of documents:
> 
> https://github.com/lervag/wiki.vim/blob/a531d999d53f043f3c11bbc7ed8aa79bc9a31b7b/autoload/wiki/tags.vim#L193
> 
> in my case, I always put tags at the middle/bottom, **which wiki.vim does not recognize**.

True.

> I can easily guess that the current implementation comes from your consideration on the speed of scanning.

Yes, that's exactly right.

> Could you think about an option to change this behavior?

Thanks for your contribution in #86. This improves the flexibility. However, it does not really allow to specify the "middle" part of a file. So I'm curious, does this suffice for you?

ompugao:
> Thanks for your contribution in #86. This improves the flexibility. However, it does not really allow to specify the "middle" part of a file. So I'm curious, does this suffice for you?

Um... Yes. On second thought, #86 is enough to me.

I have two use cases for tags (I should have summarized and put this first, sorry).
One is document-specific. The other is line-specific.
Most of my documents contains document-specific tags at the bottom of documents. #86 solves this.
On the other hand, one of my frequent line-specific tags at the middle part is "TODO" tag, which specifies today's todo with that line.
since I know what kind of line-specific tags there are in my documents, I can "vimgrep" line-specific tags, which means it is not necessary for wiki.vim to scan all line-specific tags.

lervag:
Great, happy to hear it! Also interesting to hear about your tag based workflow.

I personally use plugins like "CtrlSF" and tools like ripgrep to search for keywords in my wiki notes. This works quite well, as these tools are very fast at searching. Thus I've found that I don't really need tags, so I'm not a user of the feature myself.

PR #86, opened 2020-06-27
add option to change scanning lines for tags

ompugao:
just a sample implementation for #85


lervag:
Thanks, this seems simple and useful!

PR #88, opened 2020-07-22
feat: write_on_nav option

samgriesemer:
_**what:**_ This pull request adds a config option to automatically save files before navigating between wiki links. 

_**why:**_ This is pretty useful for me since I often instinctively follow links even when I've made local changes. In that case, I'd rather the plugin just save the file automatically and follow the link, saving me the trouble of exiting the error message and saving the file manually.

_**how:**_ The proposed solution just sticks a check for the option (`g:wiki_write_on_nav`) in the `wiki#nav#return()` and `wiki#link#open()` functions, and calls `update` if the option is set.

_**testing:**_ I've had this change in my personal wiki for some time now without any hiccups. That said, the change is pretty simple and there may be some edge cases I didn't think about, so please let me know if there any obvious issues I overlooked.

### DIFF
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -142,6 +142,9 @@ function! wiki#link#open(...) abort "{{{1
 
   try
     if has_key(l:link, 'open')
+      if g:wiki_write_on_nav
+        update
+      endif
       call call(l:link.open, a:000, l:link)
     else
       call wiki#link#toggle(l:link)
--- a/autoload/wiki/nav.vim
+++ b/autoload/wiki/nav.vim
@@ -20,6 +20,10 @@ endfunction
 " }}}1
 
 function! wiki#nav#return() abort "{{{1
+  if g:wiki_write_on_nav
+    update
+  endif
+
   if !empty(s:position_stack)
     let [l:file, l:pos] = remove(s:position_stack, -1)
     silent execute ':e ' . substitute(l:file, '\s', '\\\0', 'g')
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -430,6 +430,16 @@ OPTIONS                                                   *wiki-config-options*
 
   Default: `'~/.local/zotero'`
 
+*g:wiki_write_on_nav*
+  Option to specify whether or not to save the current file automatically
+  before navigating (forward or backward) between wiki links.
+  
+  The default behavior prompts the user to manually save any changes made to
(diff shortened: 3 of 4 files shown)


lervag:
Thanks! I'll look into it asap, but just a quick question: Are you aware of the `hidden` setting in Vim? When you write "saving me the trouble of exiting the error message and saving the file manually.", it seems to indicate that you do not `set hidden`.

samgriesemer:
You're right I didn't have it set, I wasn't aware of `hidden`. I see it makes things much nicer for navigation while keeping unsaved changes open, and I probably wouldn't have looked into this if I already had it enabled.

Even with `hidden` set, though, I think there may still be a case for the proposed option, seeing as I usually want the file to be saved when I navigate away (even if only temporarily).

nerdynewt:
Thanks for this! This saves the trouble of having to pressing enter twice on following links. But does it save the buffer though? Once I have modified multiple files and am trying to exit Vim, it shows this:
```
E37: No write since last change
E162: No write since last change for buffer "/home/vishnu/blog/_wiki/Documentation.md"
```
for each unsaved file, and takes me to each in turn (when I press ENTER), where I have to write manually.
Is this expected?

By the way, setting `hidden` only hides the buffer for me. I still get a similar message on exiting Vim, and it forces me to go through each file and save manually.

Sorry if I'm missing something obvious. I'm quite new to this.

lervag:
> You're right I didn't have it set, I wasn't aware of `hidden`. I see it makes things much nicer for navigation while keeping unsaved changes open, and I probably wouldn't have looked into this if I already had it enabled.

Yes, I think this option should be on by default.

> Even with `hidden` set, though, I think there may still be a case for the proposed option, seeing as I usually want the file to be saved when I navigate away (even if only temporarily).

Agree, I think this option is sensible.

@nerdynewt Did you actually test this with the added code and the added option? Note that this is a PR that is not yet merged, so you would have to fetch the code and checkout the specific branch. It mihgt be merged quite fast, though.

And yes, `hidden` only allows to keep hidden (and possibly modified) buffers, it does not do any kind of autosaving.

lervag:
I've merged this now. Thanks @samgriesemer for the contribution!

@nerdynewt You may test again now from the main branch of wiki.vim.

samgriesemer:
Great! Happy to contribute, even if it's only a tiny feature. Thanks for the continued work on such a great plugin.

PR #89, opened 2020-07-25
fix: relative wiki link path expansion

samgriesemer:
_**what:**_ This is a one-line addition to fix relative path name expansion when following wiki links.

_**why**_: I use relative wiki link URLs to link notes across directories. For example, I might have the notes `wiki/dir1/note1.md` and `wiki/dir2/note2.md`. To reference `note2` from within `note1` (using Markdown-style links), I would have the link `[Note 2](../dir2/note2)`. However, when visiting the link, wiki.vim leaves the target path as `wiki/dir1/../dir2/note2.md`. This seems to be okay assuming the relative path notation is interpreted properly, but I've observed unexpected behavior at times. If nothing else, Vim will show the visited file path as `wiki/dir1/../dir2/note2.md` at the bottom of the editor, which is a bit odd to look at.

_**how:**_ The fix just makes sure the `url.path` attribute is fully expanded using `fnamemodify`.

_**testing**_: The expansion resolves the observed issue in the above scenario, ensuring the visited path is `wiki/dir2/note2.md`. All paths without relative components like `..` should remain the same as before, and I've had no issues with this change in my personal wiki for the last few months.

### DIFF
--- a/autoload/wiki/url/wiki.vim
+++ b/autoload/wiki/url/wiki.vim
@@ -47,6 +47,9 @@ function! wiki#url#wiki#parse(url) abort " {{{1
     endif
   endif
 
+  " Fully expand (possibly relative) path
+  let l:url.path = fnamemodify(l:url.path, ':p')
+
   return l:url
 endfunction
 


lervag:
Hi @samgriesemer! First, again thanks for the very constructive way of addressing issues with wiki.vim! I highly appreciate the concise and clear style of your description, as well as suggesting a fix.

In this case, I think the better solution is slightly simpler. Can you test my recent update?

samgriesemer:
Appears to work perfectly! Thanks for the quick fix.

Quick question: I have one or two other PRs lined up with pretty small fixes or features (~5 LOC or so each). Do you always prefer opening a new PR even for contributions of that size? I can see where it might be undesirable to have numerous PRs each adding only a couple lines of code, when the fixes/suggestions might otherwise be suitable as an extra comment on an existing thread.

lervag:
> Appears to work perfectly! Thanks for the quick fix.

No problem!

> Quick question: I have one or two other PRs lined up with pretty small fixes or features (~5 LOC or so each). Do you always prefer opening a new PR even for contributions of that size? I can see where it might be undesirable to have numerous PRs each adding only a couple lines of code, when the fixes/suggestions might otherwise be suitable as an extra comment on an existing thread.

Feel free to open a single PR with your suggestions. I don't mind doing several things at the same time unless they are very different or significant. In general, I'm not so used to people actually suggesting fixes/improvements directly, I'm more used to issues. So I'm just happy to see real suggestions :) (Thus, also feel free to just suggest the improvements in an issue if you don't have the time for a full PR!)

samgriesemer:
That sounds good. I suppose if I have some local working code for a separate fix/feature, I'll always create a PR just to make it easier to separate, review, and merge if desired.

PR #94, opened 2020-08-30
Fix wiki-link-open-split mapping

abhineet97:
Hello.

For wiki-link-open-split, in the docs, `<c-cr>` is specified but in the actual code it had been mapped to `<c-w><cr>`.

Assuming that you're using mappings from VimWiki, I edited the mapping in the code to match that in the doc instead of the other way around.

Thanks for creating and maintaining vim.wiki!

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -141,7 +141,7 @@ function! s:init_buffer_mappings() abort " {{{1
           \ '<plug>(wiki-link-prev)' : '<s-tab>',
           \ '<plug>(wiki-link-show)' : '<leader>wll',
           \ '<plug>(wiki-link-open)' : '<cr>',
-          \ '<plug>(wiki-link-open-split)' : '<c-w><cr>',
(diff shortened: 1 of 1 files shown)


lervag:
> For wiki-link-open-split, in the docs, `<c-cr>` is specified but in the actual code it had been mapped to `<c-w><cr>`.

Ah, good catch.

> Assuming that you're using mappings from VimWiki, I edited the mapping in the code to match that in the doc instead of the other way around.

Hmm. I'm not aware of Vimwiki mappings any longer, and I didn't really aim to replicate those. I think `<c-w><cr>` is slightly better because it follows the Vim concept of prepending window commands with `<c-w>`. I don't mind using `<c-cr>` either, which could arguably be better because it is shorter...

> Thanks for creating and maintaining vim.wiki!

Happy you like it!

lervag:
I think we'll just keep the current mapping and fix the docs. Unless you should strongly disagree?

abhineet97:
Yeah, it's fine by me.

Thanks for asking!

Issue #98, opened 2020-10-13
WikiTagSearch and WikiFzFTags only searches tags in the first wiki file type


wasowski:
I was debugging a bit why the tag search mechanisms return so few tags on my tree, and found out that the search functions only  expand file names using the first type of wiki pages:

https://github.com/lervag/wiki.vim/blob/17151e0464e8aa6b3d67c8bbbdb926ee4554a945/autoload/wiki/tags.vim#L177-L180
https://github.com/lervag/wiki.vim/blob/17151e0464e8aa6b3d67c8bbbdb926ee4554a945/autoload/wiki/fzf.vim#L13-L14

The problem seems to be fetching the file extension from `g:wiki_filetypes[0]`. In my setup, I have 3 or 4 file extensions to use (for legacy reasons). Of course, I could undertake a refactoring project to uniformize these (still hope to do it one day!). For now, it seems more reasonable to make wiki.vim to search across all these extensions. Or at the very least mention this limitation in the docs for the tag search functions. 

I might use some time later this week to send you a pull request for this (right now, I need to do something else, and I wanted to see your opinion on this first anyways).

lervag:
Thanks, both for noticing, reporting and fixing! :)

PR #99, opened 2020-10-14
Extend search for tags in files with all wiki extensions (issue #98)

wasowski:
My proposal to address issue #98 . Seems to search for tags in all files.  It appears to fix the problem also for the FzF tag search, which apparently uses the same tag list.


lervag:
Agreed, this seems like a proper fix. Thanks!

Issue #100, opened 2020-10-15
WikiExport with links to markdown files


neildryan:
When using :WikiExport, any links to markdown pages (for instance `index.md`) will still be links to markdown pages in an HTML export. I was wondering if it'd be useful to change markdown links to html links? This would be useful when creating a wiki of linked HTML files so that links actually work.

If this functionality doesn't belong in this plugin, I can just make a quick find/replace function that runs before WikiExport. Thanks!

lervag:
Interesting. I don't really use the export feature for much, only to sometimes print a page and share it with people.

Could you provide a small example of this and perhaps the relevant pandoc options and "workflow" to create the relevant output?

neildryan:
Sure! Let's say there's two files `file1.md`, and `file2.md`.
```markdown
# File 1
[Link to file 2](file2.md)
```

```markdown
# File 2
```

If I convert both to HTML, File 1 will look like:
```html
<h1> File 1</h1>
<a href='file2.md'>Link to file 2</a>
```

This isn't ideal; if I'm viewing File 1 in-browser and I click on the link, I'll get `file2.md`, instead of `file2.html`. Having HTML links allows navigation in-browser between exported wiki files.

Again, I recognize that this might not be enough of the common case to justify including, though it seems like a simple find/replace on a temporary file.

lervag:
Yes, feel free to suggest a PR. I can help with the regexes, if you want. I find Vims regexes to be simpler and more convenient to work with than Pythons. And yes, I'd prefer to avoid Python.

However, I'm curious if there might be an option for pandoc that does this automatically? The reason I ask: if we do this on wiki.vim side, before exporting with pandoc, then we need to respect various settings like `g:wiki_link_extension`. I.e., the default settings rely on implicit extensions. Thus, a regex approach is not so straightforward. Still, it is not impossible, we just need to capture the corner cases.

neildryan:
[...]

PR #106, opened 2020-10-21
Link Extension Replace for WikiExport

neildryan:
First pass solution for #100, the argparse section `page.vim:300` wasn't having any effect (arguments seemed to be empty, though that seems like a config issue). Let me know what you think -- happy to make changes.

Some notes:
* This enforces a markdown or mediawiki link style syntax, but I didn't see any indication that other link formats were supported. If they are, I can add other regex or we can go in a different direction.
* This matches links that end with an extension `.md` or `.wiki`, it won't work meaningfully if `g:wiki_link_extension`is empty (this can change as well)

@lervag

### DIFF
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -308,6 +308,8 @@ function! wiki#page#export(line1, line2, ...) abort " {{{1
       let l:cfg.ext = remove(l:args, 0)
     elseif l:arg ==# '-output'
       let l:cfg.output = remove(l:args, 0)
+    elseif l:arg ==# '-link-extension-replace'
+      let l:cfg.link_replace = v:true
     elseif l:arg ==# '-view'
       let l:cfg.view = v:true
     elseif l:arg ==# '-viewer'
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -308,7 +308,7 @@ function! wiki#page#export(line1, line2, ...) abort " {{{1
       let l:cfg.ext = remove(l:args, 0)
     elseif l:arg ==# '-output'
       let l:cfg.output = remove(l:args, 0)
-    elseif l:arg ==# '-link-extension-replace'
+    elseif l:arg ==# '-link-ext-replace'
       let l:cfg.link_replace = v:true
     elseif l:arg ==# '-view'
       let l:cfg.view = v:true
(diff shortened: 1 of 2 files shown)


autoload/wiki/page.vim
```diff
@@ -470,9 +472,30 @@ function! s:export(start, end, cfg) abort " {{{1
...
+      let l:link_rx = l:link_rx . g:wiki_link_extension
+      let l:link_rx = l:link_rx . '\(#[^#\(\)\\]\{-}\)\{-})'
+      let l:pattern = '[\1](\2.html\3)'
+    elseif g:wiki_link_target_type ==# 'wiki'
+      let l:link_rx = '\[\[\([^\\\[\]]\{-}\)\.' . g:wiki_link_extension
+      let l:link_rx = l:link_rx . '\(#[^#\\\[\]]\{-}\)'
+      let l:link_rx = l:link_rx . '|\([^\[\]\\]\{-}\)\]\]'
+      let l:pattern = '\[\[\1.html\2|\3\]\]'
+    else
+      echoerr 'Unsupported link type for export replacement'
```
lervag, line 491: I agree with this approach, but I think we should be explicit about the origin. I.e., mention `g:wiki_link_target_type` in the error message.

doc/wiki.txt
```diff
@@ -662,6 +663,11 @@ the commands are also available as mappings of the form `<plug>(wiki-[name])`.
   -ext EXT~
     Specify extension to use if `fname` is not provided.
 
+  -link_ext_replace~
+    Set to true to replace in-wiki link extensions with html. This enables
+    in-browser wiki navigation. Requires`ext` be set to `html` and for
```
lervag, line 668: Missing space between `` Requires`ext` ``.

doc/wiki.txt
```diff
@@ -662,6 +663,11 @@ the commands are also available as mappings of the form `<plug>(wiki-[name])`.
...
+    `g:wiki_link_extension` to be non-empty to have any meaningful effect.
```
lervag, line 669: Change `` `g:wiki_link_extension` `` to `|g:wiki_link_extension|` to enable tag based navigation.

lervag:
Thanks! This looks quite good, I think. I had a couple of minor comments, and I want to test it locally first.

One thought: This feature might develop further, and as such, it could be nice to refactor the code into a separate function. E.g. something like this:

```vim
  " Parse wiki page content
  let l:lines = getline(a:start, a:end)

  if a:cfg.link_ext_replace
    call s:convert_links_html(l:lines)
  endif

  " ...

" And then the function something like this:
function! s:convert_links_html(lines) abort
  " Regex requires that links end with '.', even when wiki_link_extension is
  " empty.
  if g:wiki_link_target_type ==# 'md'
    let l:link_rx = '\[\([^\\\[\]]\{-}\)\](\([^\(\)\\]\{-}\)\.'
    let l:link_rx = l:link_rx . g:wiki_link_extension
    let l:link_rx = l:link_rx . '\(#[^#\(\)\\]\{-}\)\{-})'
    let l:pattern = '[\1](\2.html\3)'
  elseif g:wiki_link_target_type ==# 'wiki'
    let l:link_rx = '\[\[\([^\\\[\]]\{-}\)\.' . g:wiki_link_extension
    let l:link_rx = l:link_rx . '\(#[^#\\\[\]]\{-}\)'
    let l:link_rx = l:link_rx . '|\([^\[\]\\]\{-}\)\]\]'
    let l:pattern = '\[\[\1.html\2|\3\]\]'
  else
    echoerr 'Unsupported link type for export replacement'
    return
  endif
  call map(a:lines, 'substitute(v:val, l:link_rx, l:pattern, ''g'')')
endfunction
```

My example relies on changing the list inplace, as I think that is the most efficient approach, but one could also do it more functionally with `let l:lines = s:convert_links_html(l:lines)`.

Also, I'm curious if it is possible to reuse the regexes defined in `rx.vim`?

neildryan:
Minor changes fixed! Unfortunately, the regex in `rx.vim` don't act with any specificity to the link extension, so I don't think it'd be possible to reuse them without modification.

lervag:
Thanks! I've merged this now. There are room for improvements. For instance, wiki links like `[[name]]` do not work right now. I think it adds a new layer of complexity in that we need to substitutes. I'll leave it to you to decide if it is worth refining and improving this, or if the current version is sufficient.

neildryan:
If someone comes along and has a wiki directory without extensions and wants HTML export, we can change it then. That feels like a pretty small use case; I'll close this for now.

Issue #107, opened 2020-10-25
Page renaming fails if the target directory that does not exist


wasowski:
`wiki#page#rename` fails with a generic message `Cannot rename` if the target directory does not exist.  This makes it difficult (or more difficult than need be) to move pages around. At the very least, I think the error message should be more specific, but I am wondering whether the directories should not be created automatically with something like `mkdir -p` before attempting to move.  One possibility is to keep the current behavior of `WikiPageRename` as is (with a clearer error message) and add `WikiePageRename!` that would also create folders.

Opinions?

(I created a PR to show a test case demonstrating the problem, and also added a small patch to produce a more informative error message.)

lervag:
Thanks for opening this issue! I think it makes sense to create the directory if it does not exist (at least if we ask first). I'll look into your PR later, I've been very busy working on something on vimtex lately, and there's only so much time... :)

wasowski:
No rush at all.  I think the page renaming functionality has still several deficiencies that makes it hard to reorganize your wiki (and messy wiki is really a problem over time).  I just started with this tiny issue, because I am also very overloaded, so I am trying to move in the smallest steps possible, and slowly.  No rush at all.

lervag:
Thanks. Just as a general remark: I've found it useful to "ignore" the mess in my wiki. Entropy always increases. I try to decrease the entropy whenever I work on a topic, but I find it more important to add my thoughts with low-to-no barrier than to make everything structured and well thought-out. I consider my wiki to be very a very long-term investment and anytime I add or improve features to wiki.vim it increases the value of my wiki. [/ot]

lervag:
Ok, let's move the discussion to #108 to keep things in one thread. I'll respond more thorougly there.

PR #108, opened 2020-10-25
Allow WikiPageRename to move the page to another directory (including creating the directory)

wasowski:
These are two tests that demonstrate how WikiPageRename works for moving pages around.  It succeeds if target dir exists and fails if target dir does not exist.  These tests are supporting discussion in issue #107

### DIFF
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -95,7 +95,7 @@ function! wiki#page#rename(newname) abort "{{{1
   for l:bufname in l:bufs
     execute 'buffer' fnameescape(l:bufname)
     update
-   execute 'bwipeout' fnameescape(l:bufname)
+    execute 'bwipeout' fnameescape(l:bufname)
   endfor
   let b:wiki = l:wiki
 
@@ -127,13 +127,13 @@ function! wiki#page#rename_ask() abort "{{{1
   let l:name = input('> ')
   
   " Check if directory exists
-  let l:newpath = printf('%s/%s', expand('%:h'), l:name)
-  let l:target_dir = simplify(fnamemodify(expand(l:newpath), ':h'))
+  let l:target_dir = fnamemodify(
+    \ simplify(printf('%s/%s', expand('%:p:h'), l:name)), ":p:h")
   if !isdirectory(l:target_dir)
     redraw!
     echo "Directory '" . l:target_dir . "' does not exist. "
     if input("Create [Y]es/[n]o ? ", "Y") !=? "y"
-        return
+      return
     endif
     call mkdir(l:target_dir, "p")
   endif 
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -39,8 +39,8 @@ endfunction
 function! wiki#page#rename(newname) abort "{{{1
   redraw!
   let l:oldpath = expand('%:p')
-  let l:newpath = printf('%s/%s.%s',
-        \ expand('%:p:h'), a:newname, b:wiki.extension)
+  let l:newpath = simplify(printf('%s/%s.%s',
+        \ expand('%:p:h'), a:newname, b:wiki.extension))
 
   " Check if current file exists
   if !filereadable(l:oldpath)
@@ -95,7 +95,7 @@ function! wiki#page#rename(newname) abort "{{{1
   for l:bufname in l:bufs
     execute 'buffer' fnameescape(l:bufname)
     update
-    execute 'bwipeout' fnameescape(l:bufname)
+   execute 'bwipeout' fnameescape(l:bufname)
   endfor
   let b:wiki = l:wiki
 
@@ -120,15 +120,15 @@ function! wiki#page#rename_ask() abort "{{{1
   if input('Rename "' . expand('%:t:r') . '" [y]es/[N]o? ') !~? '^y'
     return
   endif
-  
+
   " Get new page name
   redraw!
   echo 'Enter new name (without extension):'
   let l:name = input('> ')
   
-  " Check if directory if it does not exist
-  let l:newpath = printf('%s/%s', expand('%:p:h'), l:name)
-  let l:target_dir = fnamemodify(expand(l:newpath), ':h')
+  " Check if directory exists
+  let l:newpath = printf('%s/%s', expand('%:h'), l:name)
+  let l:target_dir = simplify(fnamemodify(expand(l:newpath), ':h'))
   if !isdirectory(l:target_dir)
     redraw!
     echo "Directory '" . l:target_dir . "' does not exist. "
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -68,6 +68,14 @@ function! wiki#page#rename(newname) abort "{{{1
     return
   endif
 
+  " The target directory must exist
+  let l:target_dir = fnamemodify (expand (l:newpath), ':h')
+  if !isdirectory (l:target_dir)
+    echom 'wikiError: Cannot rename to "' . l:newpath . '".
+          \ Target directory "' . l:target_dir . '" does not exist!'
+    return
+  endif
+
   " Rename current file to l:newpath
   try
     echom printf('wiki: Renaming "%s" to "%s"',
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -68,20 +68,12 @@ function! wiki#page#rename(newname) abort "{{{1
     return
   endif
 
-  " The target directory must exist
-  let l:target_dir = fnamemodify (expand (l:newpath), ':h')
-  if !isdirectory (l:target_dir)
-    echom 'wikiError: Cannot rename to "' . l:newpath . '".
-          \ Target directory "' . l:target_dir . '" does not exist!'
-    return
-  endif
-
   " Rename current file to l:newpath
   try
     echom printf('wiki: Renaming "%s" to "%s"',
           \ expand('%:t') , fnamemodify(l:newpath, ':t'))
     if rename(l:oldpath, l:newpath) != 0
-      throw 'Cannot rename!'
+      throw 'wiki.vim: Cannot rename file!'
     end
   catch
     echom printf('wiki Error: Cannot rename "%s" to "%s"',
@@ -128,11 +120,24 @@ function! wiki#page#rename_ask() abort "{{{1
   if input('Rename "' . expand('%:t:r') . '" [y]es/[N]o? ') !~? '^y'
     return
   endif
-
+  
   " Get new page name
   redraw!
   echo 'Enter new name (without extension):'
   let l:name = input('> ')
+  
+  " Check if directory if it does not exist
+  let l:newpath = printf('%s/%s', expand('%:p:h'), l:name)
+  let l:target_dir = fnamemodify(expand(l:newpath), ':h')
+  if !isdirectory(l:target_dir)
+    redraw!
+    echo "Directory '" . l:target_dir . "' does not exist. "
+    if input("Create [Y]es/[n]o ? ", "Y") !=? "y"
+        return
+    endif
+    call mkdir(l:target_dir, "p")
+  endif 
+
   call wiki#page#rename(l:name)
 endfunction
 
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -36,11 +36,19 @@ function! wiki#page#delete() abort "{{{1
 endfunction
 
 "}}}1
-function! wiki#page#rename(newname) abort "{{{1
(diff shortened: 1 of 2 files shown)


wasowski:
I also added a proposal how to check whether a target directory exists and produce a controlled error message.  If this is what you want then we should revert the second test as success now, and start working on `WikiPageRename!` (or `WikiPageMove`?)

lervag:
From #107:

> `wiki#page#rename` fails with a generic message `Cannot rename` if the target directory does not exist. This makes it difficult (or more difficult than need be) to move pages around. At the very least, I think the error message should be more specific, but I am wondering whether the directories should not be created automatically with something like `mkdir -p` before attempting to move. One possibility is to keep the current behavior of `WikiPageRename` as is (with a clearer error message) and add `WikiePageRename!` that would also create folders.

I fully agree that we should improve this.

> I also added a proposal how to check whether a target directory exists and produce a controlled error message. If this is what you want then we should revert the second test as success now, and start working on `WikiPageRename!` (or `WikiPageMove`?).

I think a solution could be something like this:

* If the target directory does not exist, then ask the user if it should be created.

* If user responds no: abort.

* If user responds yes, then attempt to create the directory. If it fails, provide an error message.

Essentially, I think asking the user should be an OK way to handle it that should make the `!` version redundant. But I'm open to objections/different opinions, of course!

autoload/wiki/page.vim
```diff
@@ -68,6 +68,14 @@ function! wiki#page#rename(newname) abort "{{{1
     return
   endif
 
+  " The target directory must exist
+  let l:target_dir = fnamemodify (expand (l:newpath), ':h')
```
lervag, line 72: Don't add spaces before `(` in the function calls.

autoload/wiki/page.vim
```diff
@@ -68,6 +68,14 @@ function! wiki#page#rename(newname) abort "{{{1
...
+  if !isdirectory (l:target_dir)
+    echom 'wikiError: Cannot rename to "' . l:newpath . '".
```
lervag, line 74: I agree it is already better to give a specific warning. But I think we can ask the user for input on whether or not to create the directory. We can create it with `call mkdir(..., 'p')`. We can use `input()` to ask for input.

lervag: I also think we should improve the error message in line 84, e.g. `throw 'wiki.vim error: Cannot rename file!`.

test/test-page/test-move.vim
```diff
@@ -0,0 +1,16 @@
+source ../init.vim
+runtime plugin/wiki.vim
+
+silent edit wiki-tmp/BadName.wiki
+
+silent call wiki#page#rename('subdir/GoodName')
+call wiki#test#assert_equal(
+      \ expand('<sfile>:h') . '/wiki-tmp/subdir/GoodName.wiki',
+      \ expand('%:p'))
```
lervag, line 9: This test should already pass, right?

test/test-page/test-move.vim
```diff
@@ -0,0 +1,16 @@
...
+
+silent edit wiki-tmp/index.wiki
+silent call wiki#page#rename('nosubdir/index')
+call wiki#test#assert_equal(
+      \ expand('<sfile>:h') . '/wiki-tmp/nosubdir/index.wiki',
+      \ expand('%:p'))
```
lervag, line 15: And this test will fail even with your PR. For testing purposes, I think the `wiki#page#rename` function should take a "force" argument to avoid the user input, e.g. `wiki#page#rename(new_name, 1)` where the second argument is optional and if trueish implies the force.

lervag:
Since you are already working on this in your fork I'll not work on it myself yet. But I'll be glad both to help you with the PR and also, if you prefer, to finalize the implementation.

wasowski:
Thanks, I'll finish this. Just not sure when. It may take 2-3 days.

lervag:
Great. And, as you also said in #107: no rush. :)

autoload/wiki/page.vim
wasowski: This will be painful, but will comply :)

lervag: Haha :)

[ot]
That said; I find it much more important to follow local code styles when working with code than to argue my personal preferences. Of course, when I'm the main author I can decide the styles and obviously choose those that I like myself.
[/ot]

wasowski:
Sorry for beng quiet here.  I am completely swamped, but I have not forgotten this.  It just needs to wait  a bit!

lervag:
I know how it can be, no worries :)

wasowski:
I wouldn't say this is done @lervag but I think you should have a look at this as I am ending in difficulties that show my lack of understanding of wiki.vim invariants.  I think I (mostly) did what you suggested.  I ask for the permission to create a new dir (in the 'ask' function) and then the non-interactive function should just succeed.  

Still when testing I am arriving at many small problems that I am not sure if they are directly related to the patch, or whether they are problems overall that need to be looked at separately:

- The final part (`update` and `bwipeout`) has unintended side effects for me.  It closes the buffers, and if the buffer is last in a tab then it also inadvertently looses the tab (with its name and context). I am rather uncertain if this was not a problem before already.  In general I would prefer not closing the wiki buffers for updating links.  Can't we update them somehow without wipeout? reload? why `edit!` is not good enough?
- The reopenning code does not really work for me.  Renaming/moving seems to happen correctly in the file system, but the new buffer is not opened.  I was unable to figure out why? Hints?
- I noticed that renaming links fails (ie. does not rename), if simplification has changed something.  Perhaps renaming links should simplify links before comparing (or why is it a bad idea?). I did not have time to look into how link renaming works though.  I think I will look in there after I get some comments from you.

In general I struggle with the 'cwd-invariant' - vim has a working directory active in the buffer we are renaming.  This could be anything (in my tests it was often outside the wiki tree!). Typically vim file operations work with respect to the CWD, but here I would prefer to give path with respect to the position of the wiki file open, or with respect to the wiki root.  Only the former works now.  Perhaps I should implement a detection if someone starts with a slash? (or is there any machinery in place?) 

As you see these are many questions.  I think it would be best to address those that you consider bugs in this specific functionality (creating a dir) and separate the others into their own issues, to make it easier to work with in small chunks.  Comments?

lervag:
> I wouldn't say this is done @lervag but I think you should have a look at this as I am ending in difficulties that show my lack of understanding of wiki.vim invariants. I think I (mostly) did what you suggested. I ask for the permission to create a new dir (in the 'ask' function) and then the non-interactive function should just succeed.

I think this actually looks good. I'm looking into it now if it might be merged already.

> * The final part (`update` and `bwipeout`) has unintended side effects for me.  It closes the buffers, and if the buffer is last in a tab then it also inadvertently looses the tab (with its name and context). I am rather uncertain if this was not a problem before already.  In general I would prefer not closing the wiki buffers for updating links.  Can't we update them somehow without wipeout? reload? why `edit!` is not good enough?
>
> * The reopenning code does not really work for me.  Renaming/moving seems to happen correctly in the file system, but the new buffer is not opened.  I was unable to figure out why? Hints?

These looks unrelated to your patch, so I would be happy to look into them as one or more separate issues. Perhaps you could raise both of them? It would be useful with a simple example of how to observce each of them. You don't need many words for this, really, just the simple description and an example.

> * I noticed that renaming links fails (ie. does not rename), if simplification has changed something.  Perhaps renaming links should simplify links before comparing (or why is it a bad idea?). I did not have time to look into how link renaming works though.  I think I will look in there after I get some comments from you.

This is not caused by `simplify()`, no. Again, I believe this is unrelated, and feel free to post an issue with an example.

> In general I struggle with the 'cwd-invariant' - vim has a working directory active in the buffer we are renaming. This could be anything (in my tests it was often outside the wiki tree!). Typically vim file operations work with respect to the CWD, but here I would prefer to give path with respect to the position of the wiki file open, or with respect to the wiki root. Only the former works now. Perhaps I should implement a detection if someone starts with a slash? (or is there any machinery in place?)

My "philosophy" is to always work with absolute paths, because I can never assume anything about how a user "operates". I personally like to avoid keeping a distinct CWD for Vim. That is, I prefer to let CWD always be the same as the current file (`set autochdir`). I believe this might be uncommon. But both alternatives should work, and it typically does work as long as we ensure to use absolute paths in the code.

autoload/wiki/page.vim
```diff
@@ -95,7 +95,7 @@ function! wiki#page#rename(newname) abort "{{{1
   for l:bufname in l:bufs
     execute 'buffer' fnameescape(l:bufname)
     update
-    execute 'bwipeout' fnameescape(l:bufname)
+   execute 'bwipeout' fnameescape(l:bufname)
```
lervag, line 98: No need for this change?

autoload/wiki/page.vim
```diff
@@ -125,6 +125,19 @@ function! wiki#page#rename_ask() abort "{{{1
   redraw!
   echo 'Enter new name (without extension):'
   let l:name = input('> ')
+  
+  " Check if directory exists
+  let l:newpath = printf('%s/%s', expand('%:h'), l:name)
```
lervag, line 130: You should use `expand('%:p:h')` to ensure you are using an absolute path. I would also move the `simplify` from line 131 to 130. I would also consider to combine lines 130 and 131, since `l:newpath` is not used any further. E.g.:

```vim
let l:target_dir = fnamemodify(
      \ simplify(printf('%s/%s', expand('%:h'), l:name)), ':h')
```

Finally, note that the `expand()` on line 131 is not necessary.

autoload/wiki/page.vim
```diff
@@ -125,6 +125,19 @@ function! wiki#page#rename_ask() abort "{{{1
...
+  let l:target_dir = simplify(fnamemodify(expand(l:newpath), ':h'))
+  if !isdirectory(l:target_dir)
+    redraw!
+    echo "Directory '" . l:target_dir . "' does not exist. "
+    if input("Create [Y]es/[n]o ? ", "Y") !=? "y"
+        return
```
lervag, line 136: Please use two spaces per indent :)

lervag (changes requested):
The code to check if the new directory exists is only applied for `rename_ask`: I'm curious if it would be better to move this code to `rename()` and add an extra option: e.g. `ask_for_mkdir`. If true, we ask before `mkdir`, else we run `mkdir` without asking. Or something like that.

Else I think this looks good now.

wasowski:
> The code to check if the new directory exists is only applied for `rename_ask`: I'm curious if it would be better to move this code to `rename()` and add an extra option: e.g. `ask_for_mkdir`. If true, we ask before `mkdir`, else we run `mkdir` without asking. Or something like that.

hmm.  I can of course do this.  No problem. However, I thought that it was quite neat to separate a function that is interactive from the function that is non-interactive.  Perhaps I have misread the original design intention (between `wiki#page#rename` and `wiki#page#rename#ask`). Let me know, and i can of course shift the code as stated above, if you still think so.

I have addressed the other minor issues, and will look into opening separate entries for the buffer problems now.

lervag:
> > The code to check if the new directory exists is only applied for `rename_ask`: I'm curious if it would be better to move this code to `rename()` and add an extra option: e.g. `ask_for_mkdir`. If true, we ask before `mkdir`, else we run `mkdir` without asking. Or something like that.
> 
> hmm. I can of course do this. No problem. However, I thought that it was quite neat to separate a function that is interactive from the function that is non-interactive. Perhaps I have misread the original design intention (between `wiki#page#rename` and `wiki#page#rename#ask`). Let me know, and i can of course shift the code as stated above, if you still think so.

Yes, I can see what you mean. However, I'm thinking the original function should also handle the missing directories. So, as a minimum, it should raise an error or something if the target directory does not exist. My suggested solution was to move the interactive handling of a missing directory to the main function and to add an argument to avoid the interactivity. I'm thinking: If we want to change a filename and possibly move it to a different directory, then the function needs to account for this problem. If we don't explicitly say how to solve it, then we should ask. Perhaps the argument should be mandatory and be one of `[ask, abort, create]`, and default to `abort`?

> Also when renaming the page with the current patch, the path for the new file is interpreted from the position of the renamed page (relative to the renamed page). I think this is fine, but just wanted to make it clear, since you stated your 'absolute path' policy.

Yes, good point. I think this is fine. My point about absolute paths was mainly that I try and always work with paths as absolute paths in the code. Not necessarily from the user perspective.

> I have addressed the other minor issues, and will look into opening separate entries for the buffer problems now (done, issue #124).

Great! Let's merge this before working with #124, ok?

wasowski:
Moved the directory creation code to `wiki#page#rename` with an optional `dir_mode` argument as you suggested. I hope I did not mess up with re-basing and force-pushing.

lervag:
Thanks! I've merged and made some minor adjustments.

PR #110, opened 2020-10-28
Support for Asciidoctor style of links

mgoral:
This implements Asciidoctor Inter-Page Cross References style links:

```
<<file.adoc#,title>>
```

Implemented things:
- ordinary links things (creating, following, regular expresssions for parsing etc.)
- links toggling (wiki -> md -> adoc -> wiki)
- omni function (without anchors; regular expression for finding anchors is hardcoded in `get_anchors` function and creating a single regex for Asciidoc, next to `g:wiki#rx#header_items` feels hacky)

Fixes #109

PR #112, opened 2020-11-09
Fix Asciidoc's omnifunction for many links in a single line

mgoral:
In #110 I overlooked omni-completion case when there are 2 ore more links in a single line. This PR fixes it.


lervag:
Thanks!

PR #113, opened 2020-11-14
Added option to not toggle links with WikiLinkOpen

neildryan:
So, I often press enter when my cursor is over a header (`##`) or a bullet (`-` or `*`) and wiki.vim turns that WORD into a link. As `wiki#link#open` toggles the link, there isn't a suitable remapping for `<CR>`, so I added an option.

### DIFF
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -225,6 +225,11 @@ OPTIONS                                                   *wiki-config-options*
 
   Default: ''
 
+*g:wiki_link_toggle_on_open*
+  If set, transform text into links with |WikiLinkOpen|
+
+  Default: 1
+
 *g:wiki_link_target_type*
   This option may be used to pick the default style of link that will be used.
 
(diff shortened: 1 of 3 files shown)


lervag:
This looks quite good. So, if I understand correctly, you want to remove the "toggle" behaviour on pure text from the `<cr>` mapping. I assume this means you either make your links manually or with the toggle mapping (default `\wf`)?

doc/wiki.txt
```diff
@@ -225,6 +225,11 @@ OPTIONS                                                   *wiki-config-options*
 
   Default: ''
 
+*g:wiki_link_toggle_on_open*
+  If set, transform text into links with |WikiLinkOpen|
```
lervag, line 229: I would perhaps instead write:

> This option allows to disable the toggle behaviour in |WikiLinkOpen| where "normal" text is transformed into links. One may always use |WikiLinkToggle| to transform text into links.

neildryan:
Exactly, I'd like to remove the toggling from WikiLinkOpen.

lervag:
Thanks!

PR #114, opened 2020-11-18
Fix for Asciidoc's omni completion

mgoral:
Yet another one, sorry for that!


lervag:
Thanks! As I'm not really using the ascidooc style myself, I don't catch the errors either. So I'm just happy that you submit PRs for any issues you may find!

PR #141, opened 2021-04-13
Fix error on output of WikiTagSearch to scratch and cursor

nickwynja:
Running the command:

```
:WikiTagSearch -output cursor economics
```

would end up with this error:

```
Error detected while processing function wiki#tags#search[24]..<SNR>132_search[10]..<SNR>132_output_cursor:
line    3:
E119: Not enough arguments for function: wiki#link#wiki#template
E116: Invalid arguments for function add
```

This patch uses the shortened filename for the `text` argument of `wiki#link#wiki#template`.


lervag:
Thanks!

PR #142, opened 2021-04-13
fix: echo error but don't return

nickwynja:
Prior to this fix, I would get the following error when running `:WikiGraphFindBacklinks`:

```
wiki: Scanning wiki graph nodes ...
wiki: Not implemented!
Error detected while processing function wiki#graph#find_backlinks[6]..12:
line    1:
E712: Argument of get() must be a List or Dictionary
```

It looks like during the recent refactor to logging (https://github.com/lervag/wiki.vim/commit/97e8d8b26f890c5ac5ba7c97290160118536d2ea), the `gather_nodes` function returned an error even when it isn't (or doesn't appear to be, at least) a fatal error.

With this change just echoing the error doesn't block the rest of the function completing and properly shows the backlinks.

### DIFF
--- a/autoload/wiki/graph.vim
+++ b/autoload/wiki/graph.vim
@@ -171,7 +171,7 @@ function! s:gather_nodes() abort " {{{1
     endif
 
     if has_key(l:gathered, l:node)
-      return wiki#log#error('Not implemented!')
+      call wiki#log#error('Not implemented!')
     endif
 
     let l:gathered[l:node] = l:current
--- a/autoload/wiki/graph.vim
+++ b/autoload/wiki/graph.vim
@@ -170,10 +170,6 @@ function! s:gather_nodes() abort " {{{1
       endfor
     endif
 
-    if has_key(l:gathered, l:node)
-      call wiki#log#error('Not implemented!')
-    endif
-
     let l:gathered[l:node] = l:current
   endfor
 
@@ -202,7 +198,7 @@ let s:nodes = {}
 
 " }}}1
 function! s:file_to_node(file) abort " {{{1
-  return fnamemodify(a:file, ':t:r')
+  return fnamemodify(wiki#paths#shorten_relative(a:file), ':r')
 endfunction
 
 " }}}1
--- a/autoload/wiki/url.vim
+++ b/autoload/wiki/url.vim
@@ -35,7 +35,7 @@ function! wiki#url#parse(string, ...) abort " {{{1
   endif
 
   try
-    call extend(l:url, wiki#url#{l:url.scheme}#parse(l:url))
+    call extend(l:url, wiki#url#{tolower(l:url.scheme)}#parse(l:url))
   catch /E117:/
     call extend(l:url, wiki#url#generic#parse(l:url))
   endtry


lervag:
Ah, yes; I've failed at making it clear why this is "not implemented". I do not even immediately see what the reason was. I'll merge your PR, but I think it would be even better to understand which problem I have encountered here that is "not implemented". I find myself using the backlinks and graph features regularly, and it seems to generally work quite well. I assume the problem here is something that is not a big problem in my personal wiki.

lervag:
Since you are actually getting this error: are you able to see what the reason is? E.g., could you inspect the current `l:node` and the `l:gathered[l:node]` and see why there might be a clash? Can it be related to name clashes like e.g .`wiki/file.md` vs ` wiki/sub/file.md`?

nickwynja:
Ah, yes it is a filename clash in a subdirectory like you mentioned. I can push a change to warn with something like `wiki#log#warn('Duplicate filename: ' . l:node)`

lervag:
Perhaps, but it might also be better to instead fully support nested wikis instead. Instead of the current `file_to_node`, it would be better to utilize the shorten_relative function in the paths module. Could you test that on your end and inspect the result? I.e., in `s:file_to_node(...)`, use `wiki#paths#shorten_relative()` on the argument path, ala this:

```vim
return wiki#paths#shorten_relative(fnamemodify(a:file, ':r'))
```

nickwynja:
@lervag That helps! I also fixed something I was running into while the graph cache was generated. I must have some `File:` link references rather than `file:` so it was looking for a function called `wiki#url#File#parse` and throwing an error.

lervag:
Thanks. I'm curious, though: adding the `tolower` function also implies that we want to be case insensitive - why do we want that restriction. Why would you not just fix your own content and change `File:` to `file:`? My objection is not very big and probably more theoretical than realistic, but I also think it is worth commenting. So, in short: why not fix `File` to `file` on your end, instead of restricting the `wiki.vim` backend?

nickwynja:
I'd consider the scheme here similar to other URIs. [rfc3986](https://tools.ietf.org/html/rfc3986#section-3.1) mentions:

>Although schemes are case-insensitive, the canonical form is lowercase and documents that specify schemes must do so with lowercase letters.  An implementation should accept uppercase letters as equivalent to lowercase in scheme names (e.g., allow "HTTP" as well as "http") for the sake of robustness but should only produce lowercase scheme names for consistency.

`File:` showed up in my wiki from a MediaWiki import since that's the supported [syntax](https://www.mediawiki.org/wiki/Help:Images#Syntax). I think most people migrating would expect these links to work out of the box.

lervag:
> I'd consider the scheme here similar to other URIs. [rfc3986](https://tools.ietf.org/html/rfc3986#section-3.1) mentions: ...

Thanks! I was not aware of this memo/RFC, and it seems to be a very nice reference on URI's. I consider myself convinced, and I will take note of this in my own wiki :)

> `File:` showed up in my wiki from a MediaWiki import since that's the supported [syntax](https://www.mediawiki.org/wiki/Help:Images#Syntax). I think most people migrating would expect these links to work out of the box.

Agreed, and good point.

Again, thanks! I appreciate the contribution and the knowleadgable comments.

lervag:
(I forgot to tag the merge commit, but FYI: your PR was just merged.)

nickwynja:
Thanks for the merge! I very much appreciate the work you've done with wiki.vim. To me, it's exactly how plugins should work.

lervag:
Thanks! I'm not 100% satisfied myself, but I also don't have enough time to work on it. But I do think it is very close to ideal for my own purposes. Any missing feature is very non-essential, and I try to avoid implementing features that are better solved/implemented by others.

In any case, feel free to contribute further; both issues and PRs are very welcome!

PR #143, opened 2021-04-14
Create new page with fzf even if there is a match

eldridgejm:
Previously, `fzf#pages` was only capable of creating a new wiki page if the query results in no matches. This commit allows for the creation of a new page even when the query does match one or more existing pages by pressing alt-enter while in the fzf prompt.

My typical use case for `fzf#pages` is to search for a page and open it if it exists -- otherwise, create it. I often find myself in a situation where the page I'm looking for doesn't exist, but my query *does* match some existing unrelated page, and so I cannot create the new page with `fzf#pages`. For example, I'm trying to create a page named `test`, but there's a page named `technical support` (which matches due to having t,e,s,t, in that order).

This pull request uses fzf's `--expect` to allow the user to "force create" a page with the provided query as the name by pressing alt-enter instead of enter. I have left the original behavior, but in principle it could be removed to provide for a more consistent user experience. I have hardcoded the keybinding to avoid introducing more configuration, but it can be made customizable.

### DIFF
--- a/autoload/wiki/fzf.vim
+++ b/autoload/wiki/fzf.vim
@@ -17,10 +17,12 @@ function! wiki#fzf#pages() abort "{{{1
   let l:pages = globpath(l:root, l:pattern, v:false, v:true)
   call map(l:pages, '"/" . substitute(v:val, l:root . "/" , "", "")')
   call map(l:pages, {_, x -> x . "¤" . fnamemodify(x, ':r')})
+
   let fzf_opts = join([
         \ '-d"¤" --with-nth=-1 --print-query --prompt "WikiPages> "',
-        \ '--expect=alt-enter'
+        \ '--expect=' . get(g:, 'wiki_fzf_pages_force_create_key', 'alt-enter')
         \])
+
   call fzf#run(fzf#wrap({
         \ 'source': l:pages,
         \ 'sink*': funcref('s:accept_page'),
@@ -86,11 +88,7 @@ function! s:accept_page(lines) abort "{{{1
   " either empty or alt-enter, depending on if enter or alt-enter was used to
   " select, and the third line (possibly) contains the selection
   if len(a:lines) == 2 || !empty(a:lines[1])
-    let l:page = a:lines[0]
-    redraw!
-    call wiki#log#info('Opening page "' . l:page . '"')
-    sleep 1
-    call wiki#page#open(l:page)
+    call wiki#page#open(a:lines[0])
   else
     let l:file = split(a:lines[2], '¤')[0]
     execute 'edit ' . wiki#get_root() . l:file
--- a/autoload/wiki/fzf.vim
+++ b/autoload/wiki/fzf.vim
@@ -17,11 +17,14 @@ function! wiki#fzf#pages() abort "{{{1
   let l:pages = globpath(l:root, l:pattern, v:false, v:true)
   call map(l:pages, '"/" . substitute(v:val, l:root . "/" , "", "")')
   call map(l:pages, {_, x -> x . "¤" . fnamemodify(x, ':r')})
-
+  let fzf_opts = join([
+        \ '-d"¤" --with-nth=-1 --print-query --prompt "WikiPages> "',
+        \ '--expect=alt-enter'
+        \])
   call fzf#run(fzf#wrap({
         \ 'source': l:pages,
         \ 'sink*': funcref('s:accept_page'),
-        \ 'options': '-d"¤" --with-nth=-1 --print-query --prompt "WikiPages> " '
+        \ 'options': fzf_opts
         \}))
 endfunction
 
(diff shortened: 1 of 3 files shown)


autoload/wiki/fzf.vim
```diff
@@ -17,11 +17,14 @@ function! wiki#fzf#pages() abort "{{{1
   let l:pages = globpath(l:root, l:pattern, v:false, v:true)
   call map(l:pages, '"/" . substitute(v:val, l:root . "/" , "", "")')
   call map(l:pages, {_, x -> x . "¤" . fnamemodify(x, ':r')})
-
+  let fzf_opts = join([
```
lervag, line 20: I prefer to keep the empty line above to separate blocks of code. Also, although it is clearly opinionated, I've settled on the style of prefixing the `l:` to indicate local variables.

autoload/wiki/fzf.vim
```diff
@@ -78,14 +81,18 @@ endfunction
 "}}}1
 
 function! s:accept_page(lines) abort "{{{1
-  if len(a:lines) == 1
+  " if the query was so narrow that no page names matched, there will be two
+  " lines -- otherwise, three. The first line is the query, the second is
+  " either empty or alt-enter, depending on if enter or alt-enter was used to
+  " select, and the third line (possibly) contains the selection
+  if len(a:lines) == 2 || !empty(a:lines[1])
```
lervag, line 88: Thanks, I agree it is good to have a comment here to explain.

autoload/wiki/fzf.vim
```diff
@@ -78,14 +81,18 @@ endfunction
...
     let l:page = a:lines[0]
     redraw!
-    call wiki#log#info('Opening new page "' . l:page . '"')
+    call wiki#log#info('Opening page "' . l:page . '"')
```
lervag, line 91: Why did you remove the `new` here? The point was/is to indicate that it is indeed a new page, not an old, right?

Also: What happens now if we press alt-enter and the page already exists?

lervag (changes requested):
Thanks, I think these are good updates. Regarding keybinding: I assume it can not be "anything", so if it is customizable it will probably still be a choice? Also, the keybinding syntax is fzf specific, which differs from Vim notation. I'm curious, though, if alt+enter is the best choice: how about ctrl+enter? In my experience, people tend to use alt+... for global keymappings, so I would expect e.g. ctrl+enter to be less intrusive in this sense. But I'm not going to insist. Perhaps customizibility with a proper default is good here?

autoload/wiki/fzf.vim
eldridgejm: Sure thing, I'll include these style changes in the next revision.

autoload/wiki/fzf.vim
eldridgejm: If we press alt-enter and the page already exists, the existing page will be opened and "Opening page" will be displayed. This is also the reason for removing the `new `, as we are not necessarily creating a new page.

There is a decision to make as to what happens when alt-enter is pressed and `g:wiki_map_create_page` is defined. I see two natural possibilities:

1. A page whose name is the query, verbatim, is opened/created as necessary.
2. `g:wiki_map_create_page` is first applied to the query, and the resulting page name is opened/created as necessary.

The pull request currently implements the second approach. Personally, I feel it is the less surprising of the two, and the first approach would require complicating the logic.

As for the log message, I could add a check to see if the page already exists, and, if not, print that a new page is being created. One consideration here is that we'd need to calculate the effective name of the page by possibly applying `g:wiki_map_create_page`. This would involve duplicating a few lines of `wiki#page#open`. Maybe it makes sense to move the check for whether the page exists and the resulting `wiki#log#info('Opening new page')` to `wiki#page#open` itself? At first glance, that seems like a natural place for it.

eldridgejm:
Sure thing. In regards to the keybinding, I agree that making it an option and providing a good default seems to be the way to go. fzf does have a finite but [expansive](https://github.com/junegunn/fzf/blob/c7c5e7670a9bccf5c7ba2b8dd3fa73135be2b734/man/man1/fzf.1#L677 ) list of possible key combinations. Unfortunately, `ctrl-enter` is not a possibility (I believe it is not captured by some terminals). `alt-enter` was the most natural I could find, but I'm happy to change it to something else.

autoload/wiki/fzf.vim
lervag: > The pull request currently implements the second approach. Personally, I feel it is the less surprising of the two, and the first approach would require complicating the logic.

Yes, I guess you can say that; implicitly. The `g:wiki_map_create_page` is used behind the scene by `wiki#page#open`. I agree this seems to be the least surprising behaviour.

> As for the log message, ...

I agree; moving the echoing inside the `wiki#page#open` makes sense. I think it should be possible to apply the logic to the parsed url; `let l:url = wiki#url#parse()` then check the resulting `path` before `open()`.

lervag:
> Sure thing. In regards to the keybinding, I agree that making it an option and providing a good default seems to be the way to go. fzf does have a finite but [expansive](https://github.com/junegunn/fzf/blob/c7c5e7670a9bccf5c7ba2b8dd3fa73135be2b734/man/man1/fzf.1#L677) list of possible key combinations.

Thanks for the link and info; then it remains to decide the default.

> Unfortunately, `ctrl-enter` is not a possibility (I believe it is not captured by some terminals). `alt-enter` was the most natural I could find, but I'm happy to change it to something else.

Ah, yes; `enter` is often synonymous with `<c-m>`, so the "double ctrl" is often not possible. So, I think `alt-enter` is an ok choice, and if one can specify a custom key and document this, then it should be all good.

eldridgejm:
Thanks for the feedback! I have implemented all of the changes discussed. I've added a global configuration option, `g:wiki_fzf_pages_force_create_key`, which defaults to `alt-enter`. The name is verbose, so let me know if you have a better suggestion.

lervag:
Merged now. I made a minor change to the accept_page comment, but other than that this is top quality PR.  Thanks!

PR #144, opened 2021-04-15
fix: prevent duplicate tag list on reload

nickwynja:
Running `WikiTagReload` would produce a duplicate list of tag references in `WikiTagSearch`. Each time running `WikiTagReload` would add another repeated set to the search results.

This patch resets the collection to an empty list when running `WikiTagReload`.


lervag:
Thanks, good catch! (As you might realize, I don't use this feature much...)

PR #145, opened 2021-04-19
feat: add graph search to check for broken links

nickwynja:
This feature runs a link search against all files in the wiki and opens a loclist with any wiki links that point to non-existing files. It's good to check for broken links after a migration or to check if any [red links](https://en.wikipedia.org/wiki/Wikipedia:Red_link) need to be updated.

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -39,6 +39,7 @@ endfunction
 function! s:init_buffer_commands() abort " {{{1
   command! -buffer WikiCodeRun            call wiki#u#run_code_snippet()
   command! -buffer WikiGraphFindBacklinks call wiki#graph#find_backlinks()
+  command! -buffer WikiGraphCheckLinks call wiki#graph#check_links()
   command! -buffer -count=99 WikiGraphIn  call wiki#graph#in(<count>)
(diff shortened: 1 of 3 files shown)


lervag:
Nice! I was just about to shut down and go to bed, so I'll review tomorrow. But I have actually been planning this feature for a long time, but never got around to it. Thanks!

lervag:
Thanks! I've made some minor adjustments, but this was a very high quality contribution! Much appreciated!

lervag:
Also, perhaps it should be added to the docs that this command can be slow? In my wiki of ~4500 files and 209000 lines of text, it took perhaps 45 seconds.

nickwynja:
Thanks! The code base is set up nicely so that helped with getting a good PR in. I'm happy to add some documentation about the slow processing. I was also starting to look at adding caching like the backlinks graph but that seemed a bit overcomplicated since this likely wouldn't be as frequent of a command as checking backlinks.

lervag:
Yes, great, that would be appreciated. I don't think it is necessary to optimize this part, really; I would say it is something one should not need to run often, and it is not a large drawback that it can be somewhat slow. Most people probably has a much smaller set of files than myself, so it should be faster. I guess.

I've already used this now to fix ~800 broken links :)

PR #148, opened 2021-04-29
create an extension point to let users override wiki page lookups

trashhalo:
## Intent
I wish to use vim to edit my notes but my main notes editor is <https://obsidian.md>. They allow fuzzy links like `[[Bunny]]` goes to a `Bunny.md` file that can exist anywhere in your tree. This is different from wiki.vim link behavior.

## How
Allow users to define a function named `WikiSearchCommand` that takes wiki_root and fname and outputs the transformed fname.

## Example usage
In my vimrc to get wiki.vim to support obsidian links:
```vim
function! WikiSearchCommand(wiki_root, fname)
  let files = systemlist(printf('pushd %s && fd -e "md" "%s" |tail -n1 && popd', a:wiki_root, a:fname))
  return get(files, 0, a:fname)
endfunction
```

### DIFF
--- a/autoload/wiki/url/wiki.vim
+++ b/autoload/wiki/url/wiki.vim
@@ -20,6 +20,10 @@ function! wiki#url#wiki#parse(url) abort " {{{1
           \ . (l:anchors[0] =~# '/$' ? b:wiki.index_name : '')
   endif
 
+  if exists('*WikiSearchCommand')
+    let l:fname = WikiSearchCommand(g:wiki_root, l:fname)
+  endif
+
   " Extract the full path
   let l:url.path = l:fname[0] ==# '/'
         \ ? wiki#get_root() . l:fname
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -513,6 +513,13 @@ OPTIONS                                                   *wiki-config-options*
 
   Default: `'~/.local/zotero'`
 
+*WikiSearchCommand*
+  Funcref that takes wiki_root and filename and its return is the file to open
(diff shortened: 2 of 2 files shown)


lervag:
Thanks for the suggestion and the PR! My initial thought is that you are on to something that looks smart, but that I will probably want to make some adjustments before merging this. I have a busy weekend, so I will not be able to give a detailed response and review until probably Monday or Tuesday, but stay tuned. I will get back to you as soon as I get the time!

trashhalo:
> Thanks for the suggestion and the PR! My initial thought is that you are on to something that looks smart, but that I will probably want to make some adjustments before merging this. I have a busy weekend, so I will not be able to give a detailed response and review until probably Monday or Tuesday, but stay tuned. I will get back to you as soon as I get the time!

Yay! I'm not a vimscript dev soninwas just hacking my way to it working. 💯 Change everything!

lervag:
I changed this quite abit, but I think it should still work as you want it to. Try the following settings:

```vim
function! WikiSearchCommand(fname, origin)
  if empty(a:fname) | return a:origin | endif

  call wiki#paths#pushd(wiki#get_root())
  let files = systemlist(printf('fd -e "md" "%s" |tail -n1', a:fname))
  call map(files, {_, x -> fnamemodify(x, ':p')})
  call wiki#paths#popd()

  return get(files, 0, a:fname)
endfunction

let g:wiki_resolver = 'WikiSearchCommand'
```

See `:help wiki_resolver` for the help text. Note that the return value must be an absolute path. And please, let me know what you think.

PR #168, opened 2021-07-07
Check for iOS

anjiro:
Explicitly check for iOS to fix terrychou/iVim#205

### DIFF
--- a/autoload/wiki/init.vim
+++ b/autoload/wiki/init.vim
@@ -39,7 +39,7 @@ function! wiki#init#get_os() abort " {{{1
   if has('win32')
     return 'win'
   elseif has('unix')
-    if has{'ios') || system('uname') =~# 'Darwin'
+    if has('ios') || system('uname') =~# 'Darwin'
       return 'mac'
     else
       return 'linux'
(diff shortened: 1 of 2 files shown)


autoload/wiki/init.vim
```diff
@@ -39,7 +39,7 @@ function! wiki#init#get_os() abort " {{{1
   if has('win32')
     return 'win'
   elseif has('unix')
-    if system('uname') =~# 'Darwin'
+    if has{'ios') || system('uname') =~# 'Darwin'
```
lervag, line 42: It seems you've used `{` instead of `(` here. Also, perhaps we should also add `has('mac')`?

autoload/wiki/u.vim
```diff
@@ -48,7 +48,7 @@ function! wiki#u#get_os() abort " {{{1
   if has('win32') || has('win32unix')
     return 'win'
   elseif has('unix')
-    if has('mac') || system('uname') =~# 'Darwin'
+    if has('mac') || has{'ios') || system('uname') =~# 'Darwin'
```
lervag, line 51: `{` should be `(`

lervag:
Thanks; I'll merge this after you address my two comments.

anjiro:
Whoops, tiny font on iPhone screen leads to typos. Fixed!

lervag:
Thanks!

PR #169, opened 2021-07-09
Fix error in :WikiPageRename

digital-carver:
Trying to run `:WikiPageRename` asks the confirmation prompt and then the new name, but errors out after that: 
```

Error detected while processing function wiki#page#rename_ask[11]..wiki#page#rename:
line   64:
E119: Not enough arguments for function: bufnr
E15: Invalid expression: bufnr()

```

I believe the problem is the empty call to `bufnr()` at lines 102 and 118 of ` autoload/wiki/page.vim `. Changing those to `bufnr('')` (like in scratch.vim) fixes the issue.

### DIFF
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -99,7 +99,7 @@ function! wiki#page#rename(newname, ...) abort "{{{1
   endif
 
   " Rename current file to l:newpath
-  let l:bufnr = bufnr()
+  let l:bufnr = bufnr('')
   try
     call wiki#log#info(
           \ printf('wiki: Renaming "%s" to "%s" ...',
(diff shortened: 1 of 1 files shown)


lervag:
I'm curious; which version of Vim are you on? On my end, this fix is not necessary. I use Vim 8.2 and neovim 0.5.

lervag:
I'm merging as I don't see a reason why not, but I'm still curious :)

digital-carver:
> I'm curious; which version of Vim are you on? On my end, this fix is not necessary. I use Vim 8.2 and neovim 0.5.

Ah, I did not realize it was a version-dependent issue. I was on Vim 8.0.1568, which seems to be the latest version in my distro's official repository. I've found a different repo with more recent updates though, and am updating to Vim 8.2 right now. Thanks for indulging your curiosity and prompting this update. :)

lervag:
I guess there's been a change to the `bufnr()` function since Vim 8.0, then. And no problem; I usually recommend people to try and use the latest versions of Vim (or neovim). Updates are usually improvements.

PR #173, opened 2021-07-18
Support default 'date' on nvim 0.5 on OSX

aignas:
Currently attempting to create a monthly summary by pressing `<leader>wm` leads to an error:
```
Error detected while processing function wiki#journal#freq[16]..wiki#date#format[1]..<SNR>92_date:
line    5:
E766: Insufficient arguments for printf()
```

This attempts to work around this by passing the format of the date  as a parameter to `printf` such that it is not attempting to interpret the date format as parameters to `printf` itself.


lervag:
Thanks!

Issue #104, opened 2020-10-17
Using Funcrefs/lambdas with wiki_map_link_create


savq:
# Description:

Currently wiki.vim doesn't support passing funcrefs/lambdas to the `wiki_map_link_create` option (because of strange VimL behavior).

# Minimal working example

If we create a lambda: `let CreateLoudLinks = {x -> toupper(x) . '!'}`, then:

This works

```
echo exists('*CreateLoudLinks')             " string
" = 1

echo call(CreateLoudLinks, ['hello world']) " symbol
" = HELLO WORLD!
```

But this doesn't work

```
echo exists(CreateLoudLinks)                   " symbol
" = E729: using Funcref as a String

echo call('*CreateLoudLinks', ['hello world']) " string
" = E117: Unknown function: CreateLinkLinks
```

# The problem

The `exists` function always expects a string, however, the `call` function will fail if it receives a funcref as a string (but not as a symbol).

# Suggested solution

In the source code the relevant bit seems to be:

```
" wiki.vim/autoload/wiki/link/word.vim

if !empty(g:wiki_map_link_create) && exists('*' . g:wiki_map_link_create)
  let l:url = call(g:wiki_map_link_create, [a:url])
endif
```

Allowing the user to pass a funcref, and typecheck `wiki_map_link_create` here, before calling `exists` and `call` seems to me like a reasonable workaround for the inconsistency between the two functions.

savq:
To add more context, the reason I'm trying to pass a funcref rather than the name of a function is because I'm calling a lua function, and having another function just to wrap it seems a bit cluttered. I'd rather wrap it with a lambda.

lervag:
I agree this would be nice, but unfortunately, it is not possible. The reason is that the option names are all lowercase. Function names must start with a capital letter, see `:help E704`.

Note, this should really be just a minor inconvenience, i.e. that you need two lines instead of one, e.g.:

```vim
" We can't do this:
let g:wiki_map_link_create = {x -> toupper(x) . '!'}

" But this works
let CreateLoudLinks = {x -> toupper(x) . '!'}
let g:wiki_map_link_create = 'CreateLoudLinks'
```

savq:
Yes, you're right. I made sure to name the lambda in the example because passing it directly wouldn't work.

[...]

PR #174, opened 2021-07-18
Allow funcrefs in `wiki_map_link_create`

savq:
Funcrefs must start with a capital letter, so this only really works
with Lua functions in Neovim.

See #104


lervag:
Thanks! Interesting to learn that Lua does not have the same limitation on the variable name.

PR #179, opened 2021-07-24
Add tags to omnicompletion

anjiro:
It’s got a bit of a hack to avoid matching the end-of-tag marker when in `findstart` mode, but it seems to work in my basic testing. It does depend on `g:wiki_tags_format_pattern` having a `\ze`, but since that’s the default and how it’s documented it should probably be ok.

### DIFF
--- a/autoload/wiki/complete.vim
+++ b/autoload/wiki/complete.vim
@@ -189,6 +189,30 @@ function! s:completer_zotero.complete(regex) dict abort " {{{2
 endfunction
 
 " }}}1
+" {{{1 Tags
+
+let s:completer_tags = {}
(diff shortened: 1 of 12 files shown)


lervag:
I'll be happy to consider this PR, but please: the PR itself contains commits already merged. Can you please rebase your work on top of the current master and only propose _new_ commits? A simple way to do it is to fork the latest master, then just copy your files and commit them either all in one or with some logical commits - feel free to ask if you want more help with it.

anjiro:
Yes, will do that.

lervag:
#180 replaces this PR.

PR #180, opened 2021-07-24
Add tag completion

anjiro:
It’s got a bit of a hack to avoid matching the end-of-tag marker when in `findstart` mode, but it seems to work in my basic testing. It does depend on `g:wiki_tags_format_pattern` having a `\ze`, but since that’s in the default pattern and is featured in the documentation examples, it seems likely to be fine.

### DIFF
--- a/autoload/wiki/complete.vim
+++ b/autoload/wiki/complete.vim
@@ -189,6 +189,41 @@ function! s:completer_zotero.complete(regex) dict abort " {{{2
 endfunction
 
 " }}}1
+" {{{1 Tags
+
+let s:completer_tags = {}
+
+function! s:completer_tags.findstart(line) dict abort " {{{2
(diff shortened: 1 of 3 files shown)


lervag:
Thanks! The code looks good. It would be nice if you could also add:

* Some documentation (in doc/wiki.txt in the completion section).
* Either one or more simple tests (see `test/test-complete/test-wikilinks.vim` for an example), or as a minimum some example files in `test/wiki-basic` or similar, which I can use to add the tests myself.

anjiro:
Added docs, will look at tests in the next few days.

anjiro:
I added some extra tags to  to `tagged.wiki` to enable basic testing. I’m doing almost all of my work with iVim on iOS at the moment so implementing a full test is a bit too complex.

lervag:
Thanks! I've added test and made some minor modifications.

PR #182, opened 2021-07-27
Clearfication of substitution functions

Ginner:
I've tried expanding on the function substitution by supplying the predefined function as an example. Initially the example replaced the 'Pre-defined functions' section. However, I think this should remain.


lervag:
Thanks! I've merged this now. I also changed the text for a better example structure.

PR #185, opened 2021-07-31
Proposed cearification in the docs

Ginner:
Added a description of the predefined template substitution function.


lervag:
Thanks!

PR #186, opened 2021-08-16
Add bold+italic matching

anjiro:
Modified `wiki#rx#surrounded()` to reverse `a:chars`, which I think is what it was supposed to do anyway, and added patterns for bold+italic and italic+bold.


lervag:
Thanks!

PR #196, opened 2021-11-04
Restore explicit check for iOS

anjiro:
#168 fixed terrychou/iVim#205 due to system calls not working during startup; 1459952e92b58b7682c7513338ef47febca5495e regresses back to this bug. This restores the explicit check for iOS to avoid the problem.


lervag:
Thanks! I removed it because I couldn't find it documented. Perhaps it is part of MacVim, which seems to have documentation that is not part of Vim or neovim?

anjiro:
The iVim documentation doesn’t include it in `:h feature-list`, so I’m guessing it’s just missing documentation.

PR #197, opened 2021-11-05
Add option for case-insensitive omnicomplete

anjiro:
My wiki files all have the first letter capitalized, which makes case-sensitive searching inconvenient. I added an option `g:wiki_completion_case_sensitive` to optionally do case-insensitive searching.

### DIFF
--- a/autoload/wiki/complete.vim
+++ b/autoload/wiki/complete.vim
@@ -29,8 +29,7 @@ function! wiki#complete#link(lead, line, pos) abort " {{{1
   let l:candidates = s:completer_wikilink.complete(l:input)
   call map(l:candidates, 'v:val.word')
   if !s:completer_wikilink.is_anchor
-    return filter(l:candidates, {_,x -> match(x,
-          \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . l:input) == 0})
+    return s:filter_candidates(l:candidates, '^' . l:input)
   endif
 
   return map(l:candidates, 'l:base . v:val')
@@ -100,13 +99,10 @@ function! s:completer_wikilink.complete_anchor(regex) dict abort " {{{2
   let l:length = strlen(l:base)
 
   let l:anchors = wiki#toc#gather_anchors(l:url)
-  call filter(l:anchors, {_,x -> match(x,
-        \ (g:wiki_completion_case_sensitive ? '\C' : '\c')
-        \ . '^' . wiki#u#escape(l:base) . '[^#]*$') >= 0})
+  call s:filter_candidates(l:anchors, '^' . wiki#u#escape(l:base) . '[^#]*$')
   call map(l:anchors, 'strpart(v:val, l:length)')
   if !empty(a:regex)
-    call filter(l:anchors, {_,x -> match(x,
-        \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . a:regex) >= 0})
+    call s:filter_candidates(l:anchors, a:regex)
   endif
 
   return l:anchors
@@ -126,8 +122,7 @@ function! s:completer_wikilink.complete_page(regex) dict abort " {{{2
         \ empty(b:wiki.link_extension)
         \ ? 'l:pre . fnamemodify(v:val, '':r'')'
         \ : 'l:pre . v:val')
-  call filter(l:cands, {_,x -> match(x,
-        \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . a:regex) >= 0})
+  call s:filter_candidates(l:cands, a:regex)
 
   call sort(l:cands)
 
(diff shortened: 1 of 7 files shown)


autoload/wiki/complete.vim
```diff
@@ -29,7 +29,8 @@ function! wiki#complete#link(lead, line, pos) abort " {{{1
   let l:candidates = s:completer_wikilink.complete(l:input)
   call map(l:candidates, 'v:val.word')
   if !s:completer_wikilink.is_anchor
-    return filter(l:candidates, 'stridx(v:val, l:input) == 0')
+    return filter(l:candidates, {_,x -> match(x,
+          \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . l:input) == 0})
```
lervag, line 33: This is _probably_ OK, but note that regex matching is usually much slower than the `stridx`. One possibility here would be to only do regex matching if we are case insensitive. On the other hand, premature optimization is not really necessary...

autoload/wiki/complete.vim
```diff
@@ -99,10 +100,13 @@ function! s:completer_wikilink.complete_anchor(regex) dict abort " {{{2
...
   let l:anchors = wiki#toc#gather_anchors(l:url)
-  call filter(l:anchors, 'v:val =~# ''^'' . wiki#u#escape(l:base) . ''[^#]*$''')
+  call filter(l:anchors, {_,x -> match(x,
+        \ (g:wiki_completion_case_sensitive ? '\C' : '\c')
+        \ . '^' . wiki#u#escape(l:base) . '[^#]*$') >= 0})
   call map(l:anchors, 'strpart(v:val, l:length)')
   if !empty(a:regex)
-    call filter(l:anchors, 'v:val =~# ''' . a:regex . '''')
+    call filter(l:anchors, {_,x -> match(x,
+        \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . a:regex) >= 0})
```
lervag, line 109: I think here and at line 103 it would be better to swap `=~#` with `=~?` for the case sensitive stuff. E.g.:

```vim
     call filter(l:anchors, printf('v:val =~%s ''%s''',
           \ g:wiki_completion_case_sensitive ? '#' : '?',
           \ a:regex))
```

autoload/wiki/complete.vim
```diff
@@ -207,7 +212,8 @@ endfunction
 
 function! s:completer_tags.complete(regex) dict abort " {{{2
   let l:candidates = keys(wiki#tags#get_all())
-  call filter(l:candidates, {_, x -> stridx(x, a:regex) >= 0})
+  call filter(l:candidates, {_, x -> match(x,
+        \ (g:wiki_completion_case_sensitive ? '\C' : '\c') . a:regex) >= 0})
```
lervag, line 216: I notice the pattern is repeated several times. Perhaps it would look better if we abstracted into a separate function somehow? E.g. `call s:filter(l:candidates, a:regex)` where `s:filter` does the necessary work? This could also help slightly improve things by "caching" the `g:wiki_completion_case_sensitive` choice (sorry, premature optimization again; I can't help it).

lervag: Caching the case sensitive choice would break some of your tests, though, so I would probably not care about that. But simplifying the pattern by abstracting it to a separate filter function seems possibly useful to me.

lervag (commented):
Thanks! Again I'm glad to see your thoughtful and high quality contribution. I do have some comments, but they are mostly neutral. I'd like to hear your thoughts, but we might conclude to merge without any further updates.

autoload/wiki/complete.vim
anjiro: Just for fun, I threw together a test of `stridx` vs `match` (see [this gist](https://gist.github.com/anjiro/db5b539242d6791c46c59fb6025b9874)). It generates a list with 100,000 100-character strings, then calls `filter` with a random 10-character string, using `stridx` or `match`. `stridx` is indeed faster than `match` - by 0.25s on my iPad, and by .19s on my MacBook Pro. So I think we’re safe!

autoload/wiki/complete.vim
anjiro: Any reason to prefer a string over a lambda? I tend to like the lambas better for readability - the excessiv number of single-quotes necessary bothers me.

autoload/wiki/complete.vim
anjiro: I think the only place that would break down is on line 32 where the comparison is `== 0` rather than `>= 0`. I’ll give a try breaking it out into a function for the other lines and add a commit if it seems good.

I’d argue in general against caching `g:` options; from a usability standpoint, it’s better to allow an option to change whenever the user wants to, rather than requiring a restart.

autoload/wiki/complete.vim
lervag: Thanks! for testing - it confirms my understanding, but it also shows that the difference is insignificant.

autoload/wiki/complete.vim
lervag: No, not really; if it is a more or less trivial string I often prefer it, but when it becomes more complex, I agree that lambdas are usually much clearer.

autoload/wiki/complete.vim
```diff
@@ -29,7 +29,7 @@ function! wiki#complete#link(lead, line, pos) abort " {{{1
   let l:candidates = s:completer_wikilink.complete(l:input)
   call map(l:candidates, 'v:val.word')
   if !s:completer_wikilink.is_anchor
-    return filter(l:candidates, 'stridx(v:val, l:input) == 0')
+    return s:filter_candidates(l:candidates, '^' . l:input)
```
lervag, line 32: Ah, that's smart!

lervag:
Thanks, I've merged now!

PR #198, opened 2021-11-08
Use wiki_index_name global to set a custom index page name.

MosaicMan:
Love the work you have put into Wiki.vim.

This is a small change to configure the name of the index page, by using the wiki_index_name variable in the goto_index function.

Thanks for considering it.

### DIFF
--- a/autoload/wiki.vim
+++ b/autoload/wiki.vim
@@ -54,7 +54,8 @@ endfunction
 " }}}1
 
 function! wiki#goto_index() abort " {{{1
-  call wiki#url#parse('wiki:/index').follow()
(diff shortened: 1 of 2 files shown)


lervag:
> Love the work you have put into Wiki.vim.

Thanks! :D

> This is a small change to configure the name of the index page, by using the wiki_index_name variable in the goto_index function.

Yes, that's a natural option. I'll review and probably merge immediately. Thanks!

lervag:
I see I already added that option earlier, but I forgot to 1) document it, and 2) use it everywhere. I think your PR solved both issues. Thanks!

MosaicMan:
Awesome. Thanks again!

PR #201, opened 2021-11-11
Remove unneeded bolditalic and italicbold regexes

anjiro:
Due to improved wiki-ft syntax matching; see lervag/wiki-ft.vim#8 .


lervag:
Thanks!

PR #204, opened 2021-11-16
Added command to follow link in new tab

endter:
Created command WikiLinkFollowTab to open the link under cursor in a new tab. No default keybinding is associated. This behavior will be familiar to those used to navigating Zettelkästen via GUI applications, and there is no obvious drawback to including it in the project.


lervag:
Thanks! It would make sense to also add a corresponding plug mapping and documentation, see e.g. `:help WikiLinkFollowSplit` for an example.

endter:
Done! I've mapped wiki-link-follow-tab to <c-w>o (as in "open") to be consistent with the mapping of wiki-link-follow-split. I have updated the documentation. 

My preference would be for <leader><cr> as the mapping, but it is likely to create conflicts.

Sorry for the double-commit. I am new to both vim and git.

endter:
Done! I've mapped wiki-link-follow-tab to c-w o (as in "open") to be consistent with the mapping of wiki-link-follow-split. I have updated the documentation.

My preference would be for as the mapping, but it is likely to create conflicts.

Sorry for the double-commit. I am new to both vim and git.

endter:
Done! This mapping is consistent with the WikiLinkFollowSplit mapping.

lervag:
Thanks!

PR #208, opened 2021-12-22
Make goto_index obey wiki_link_extension

krisgry:
I experience the following issue, which I believe to have fixed:
I have `let wiki_link_extension='.wiki'`. When I have a markdown file open (from random editing) and want to open the wiki index (pressing <leader>ww), I'm brought to `index.**md**` instead of `index.wiki` (both files happen to be in my root folder).  

The following fix adds g:wiki_link_extension to g:wiki_index_name.

Thanks for a great plugin btw!

### DIFF
--- a/autoload/wiki.vim
+++ b/autoload/wiki.vim
@@ -54,7 +54,7 @@ endfunction
 " }}}1
 
 function! wiki#goto_index() abort " {{{1
-  let l:index_url = printf('wiki:/%s', g:wiki_index_name)
+  let l:index_url = printf('wiki:/%s', g:wiki_index_name . g:wiki_link_extension)
   call wiki#url#parse(l:index_url).follow()
 endfunction
 


lervag:
> I experience the following issue, which I believe to have fixed: I have `let wiki_link_extension='.wiki'`. When I have a markdown file open (from random editing) and want to open the wiki index (pressing ww), I'm brought to `index.**md**` instead of `index.wiki` (both files happen to be in my root folder).

This is slightly more complicated, I think. One idea with wiki.vim is that it should "just work". You could have multiple wikis; the `g:wiki_root` would specify a main wiki, but if you open a file within a different project, then you would expect the `:WikiIndex` to go to the current projects index. And the current project may use different extensions. Your fix would then break this, I believe.

Still, I see your point, and perhaps my comment is to principled. I.e., this PR will fix your problem, and I don't think it would break my personal work flow. So a pragmatic choice might be to accept this and instead react to new issues if this would break anyone elses workflow...

> Thanks for a great plugin btw!

Happy you like it! :)

krisgry:
Yea, my fix was rather easy so I thought there might be a "philosophical" reason :) Completely up to you what you decide to do.

How does `:WikiIndex` find the index then, does it consider the wiki graph? (I interpret your comment to say that it does not need `g:wiki_root`) Perhaps I can refine my workflow here :smiley:

lervag:
> How does `:WikiIndex` find the index then, does it consider the wiki graph?
> (I interpret your comment to say that it does not need `g:wiki_root`)

`:WikiIndex` calls `wiki#goto_index` which uses `wiki#url#parse` on the wiki scheme link `/{g:wiki_index_name}`. By default, the `wiki#url#wiki#resolver` is used to resolve the link path. It is not fully trivial; but the idea is that the initial `/` will be replaced by the output of `wiki#get_root()`. This function will first look for a local wiki root. If it can't find it, it will resolve to `g:wiki_root`. Finally, also notice that the resolver will try and find the proper extension:

https://github.com/lervag/wiki.vim/blob/e23358368e3db6408abec9d0b7d2be6b78a9c1fc/autoload/wiki/url/wiki.vim#L41-L58

I hope the above may be helpful to explain how things work. Feel free to ask follow-up questions. And so:

> Yea, my fix was rather easy so I thought there might be a "philosophical" reason :) Completely up to you what you decide to do.

I believe that it may be better to improve the resolver wrt. the index extension thing. And so, to be more clear: I believe it would be good to create some concrete examples of what you are trying to do and how you expect wiki.vim to behave. You write:

> When I have a markdown file open (from random editing) and want to open the wiki index (pressing ww), I'm brought to index.**md** instead of index.wiki (both files happen to be in my root folder).

Can you explain this in more depth? Preferably with concrete, minimal examples.

krisgry:
Sorry for the delayed response. Time has not been in abundance, and will not be in the foreseeable future, so I will (temporarily?) close this PR.

Thanks for your explainations. I think my confusion is related to the scenario of having multiple wikis, which I had not considered as a use case. I will look at the resolver if I get back to this problem. 

As for an in-depth explanation of my the behavior I saw vs my expectations.

0. Have both index.wiki and index.md in wiki_root. Below are my wiki.vim settings
1. I am editing some markdown file: `vim test.md` 
2. I realize that I should find/add something in my personal wiki, so I head over to the wiki root: `\ww` 
3. I now find myself in index.md, but I would expect to be in index.wiki since I have `g:wiki_link_extension` set to wiki.

I do not expect you to take time to follow up on this, since I do not have time myself, and I am fine with the above (I can simply remove index.md from my wiki folder, which was only used for initial testing). I submitted this PR thinking that it might have been an edge case that had not been considered (since I believe that not many users have index.md and index.wiki). Once again, thanks for this plugin, and your general vim inspiration :)

```
let g:wiki_root = '~/Documents/doc/wiki'
let g:wiki_link_extension = '.wiki'
let g:wiki_map_link_create = 'NoWhitespaceLinks'
let g:wiki_link_target_type = 'md'

function NoWhitespaceLinks(text) abort
  return substitute(tolower(a:text), '\s\+', '_', 'g')
endfunction

let g:wiki_date_exe = 'gdate' 
let g:wiki_list_todos = ['TODO','DOING','DONE'] "list to toggle through with wiki-list-toggle

"pandoc setup for export
let g:wiki_export = {
    \ 'args' : '--self-contained --bibliography=<path>.bib --template=GitHub.html5',
    \ 'from_format' : 'markdown',
    \ 'ext' : 'html',
    \ 'view' : v:true,
    \ 'link_ext_replace': v:true, 
    \ 'output': fnamemodify(tempname(), ':h'),
    \}

if g:env =~ 'DARWIN'
  let g:wiki_viewer = {
      \ '_' : 'open',
      \}
elseif g:env =~ 'LINUX'
  let g:wiki_viewer = {
      \ '_' : 'xdg-open',
      \}
endif
let g:wiki_filetypes = ['wiki', 'md']
let g:wiki_write_on_nav = 0

let g:wiki_toc_depth = 2
let g:wiki_file_open = 'personal#wiki#file_open'
let g:wiki_completion_case_sensitive = 0
```

lervag:
> Sorry for the delayed response. Time has not been in abundance, and will not be in the foreseeable future, so I will (temporarily?) close this PR.

No problem!

> As for an in-depth explanation of my the behavior I saw vs my expectations.
>
>   0. Have both index.wiki and index.md in wiki_root.
> 
> ...
> 
>   3. I now find myself in index.md, but I would expect to be in index.wiki since I have `g:wiki_link_extension` set to wiki.

I tested, and I can reproduce your problem even without `index.md` in my root. I agree this is a bug, and I'll look into it!

lervag:
I believe this is fixed now. Please test on your end.

Thanks for reporting; this has now resolved a slight annoyance for myself as well :)

krisgry:
Glad I was able to communicate the problem :) I have tested with the master branch, which includes the commit linked above, but I still see the same problem (index.md is opened when pressing \ww from test.md)

lervag:
Sorry for the late reply! I've tried to replicate your problem, but I can't. I have these files:

```
❯ tree
. (TEMPORARY PATH)
├── test.vim
└── wiki
    ├── index.md
    └── index.wiki
```

The contents of the `index.md` and `index.wiki` files are simply "hello markdown" and "hello wiki" (or similar). The `test.vim` looks like this:

```vim
set nocompatible
set runtimepath^=~/.local/plugged/wiki.vim
filetype plugin indent on
syntax enable

let g:wiki_root = fnamemodify("wiki", ":p")
let g:wiki_filetypes = ["wiki", "md"]
let g:wiki_link_extension = ".wiki"
let g:wiki_link_target_type = 'md'

runtime plugin/wiki.vim

WikiIndex
```

I start neovim with `nvim --clean -u test.vim`, and as expected, it will bring me to `index.wiki`.

Could you try to adjust the example to make a reproducible example that shows what is still not working as expected on your end?

lervag:
I think it is fixed now, please test.

lervag:
Glad to hear it! :)

PR #210, opened 2021-12-31
Fix typos in doc file

agorf:


doc/wiki.txt
```diff
@@ -167,29 +167,27 @@ GUIDE                                                        *wiki-intro-guide*
 
 Executive summary~
 * Speed is key - make sure you have a fast workflow.
-  * Use a global keyboard shortcut to open the wiki index, e.g. <alt-n>.
+  * Use a global keyboard shortcut to map |:WikiIndex| to open the wiki index,
```
lervag, line 170: No, that's not really what I mean. When I say _global_ I don't mean a Vim mapping, but a mapping at the window/desktop manager level. I believe this is more clear in the following text, but I can see that it is unclear here.

doc/wiki.txt
```diff
@@ -167,29 +167,27 @@ GUIDE                                                        *wiki-intro-guide*
...
+    e.g. <alt-n>.
   * Use |:WikiFzfPages| and `ripgrep` (with e.g. ctrlsf.vim) for searching.
 * Write short pages and use links between related pages.
 * Don't obsess about the wiki structure.
 
-
```
lervag, line 175: The idea behind this and the next extra line spacing is to set off the "preface" text. The real guide/essay starts at "People write things down...". Perhaps I should set it off in a different manner?

doc/wiki.txt
agorf: Oops. Yeah, I never realized you are referring to the wm level, but it makes sense. Maybe it needs some clarification as "global" can be interpreted in many ways (I thought not at the buffer level)

Reverted with 5f386b4936120ac918b244bd3c42a74db61cde64

doc/wiki.txt
```diff
@@ -322,8 +320,8 @@ top of a local folder of plain text Markdown files.
 ==============================================================================
 CONFIGURATION                                                     *wiki-config*
 
-The following subsections presents a list of available options and events that
-can be used to customize the behaviour of wiki.vim.
+The following subsections present a list of available options and events that
+can be used to customize the behavior of wiki.vim.
```
lervag, line 323: Re: behavior - I've aimed for UK english, so should it not be "behaviour"?

doc/wiki.txt
agorf: Oh, I thought it was forgotten. Maybe use some kind of separator?

lervag (changes requested):
Thanks! I really appreciate this! I had some minor comments, so let's settle them first.

doc/wiki.txt
lervag: I'll fix in an update.

doc/wiki.txt
agorf: True, you repeat `behaviour` many times later as well. My bad.

f0956a018c2d46dd5d47fb03b2a2f16fc6c122a0

doc/wiki.txt
agorf: I'm reverting for now.

agorf: 9af22a3fa3694ce186bbcdb2b2017a4ff2fca956

lervag:
Thanks! I've merged and made the suggested updates from my side as well. I also improved the paragraph on the global keyboard shortcut; or at least, I tried to improve it :)

agorf:
Awesome. I think it's more clear now. Thanks for wiki.vim and happy new year!

lervag:
Great; thanks for contributing! And happy new year to you too! Let's hope for a less pandemic 2022 :)

PR #211, opened 2022-01-05
Suggest having a wm shortcut for WikiJournal

agorf:
Hello. I've found this very helpful as well (actually I'm using it more than `WikiIndex`). Feel free to close if you think it's not.

Thanks for wiki.vim!


lervag:
Thanks, I agree this is a good idea and I use a similar shortcut as well. :)

lervag:
I just pushed a minor adjustment to make it more "personalized". I hope you don't mind!

PR #216, opened 2022-01-28
Make it possible to customize local journal mappings

aignas:
A small addition of an extra setting and a short intro in the docs.

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -176,7 +176,7 @@ function! s:init_buffer_mappings() abort " {{{1
 
   call extend(l:mappings, get(g:, 'wiki_mappings_local', {}))
   if b:wiki.in_journal
-    call extend(l:mappings, get(g:, 'wiki_mappings_journal_local', {}))
+    call extend(l:mappings, get(g:, 'wiki_mappings_local_journal', {}))
   endif
 
   call wiki#init#apply_mappings_from_dict(l:mappings, '<buffer>')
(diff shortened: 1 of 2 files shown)


autoload/wiki/buffer.vim
```diff
@@ -175,6 +175,9 @@ function! s:init_buffer_mappings() abort " {{{1
   endif
 
   call extend(l:mappings, get(g:, 'wiki_mappings_local', {}))
+  if b:wiki.in_journal
+    call extend(l:mappings, get(g:, 'wiki_mappings_journal_local', {}))
```
lervag, line 179: Could I suggest `local_journal` instead of `journal_local`?

doc/wiki.txt
```diff
@@ -616,6 +617,9 @@ OPTIONS                                                   *wiki-config-options*
   visual mode, respectively. The available `<plug>` mappings are listed in
   |wiki-mappings|.
 
+  For any local mappings that should only be applied in journal entries,
+  consider setting |g:wiki_mappings_journal_local|.
+
```
lervag, line 622: Perhaps one should add an example for this as well?

lervag (changes requested):
Thanks! This looks like a useful addition. I have two minor comments.

doc/wiki.txt
aignas: Added an example and an extra table for the defaults.

autoload/wiki/buffer.vim
aignas: Done

lervag:
Thanks! I hope you don't mind that I made a minor adjustment on top of this. In any case, I appreciate your contribution!

aignas:
> Thanks! I hope you don't mind that I made a minor adjustment on top of this. In any case, I appreciate your contribution!

Thanks! no worries. My main goal was to get the setting in, so I am happy that you could review this this quickly!

PR #221, opened 2022-04-05
Support funcrefs for g:wiki_map_create_page

jgilchrist:
They are supported for the other wiki_map* callbacks.

### DIFF
--- a/autoload/wiki/page.vim
+++ b/autoload/wiki/page.vim
@@ -6,7 +6,7 @@
 
 function! wiki#page#open(page) abort "{{{1
   let l:page =
-        \ !empty(g:wiki_map_create_page) && exists('*' . g:wiki_map_create_page)
+        \ !empty(g:wiki_map_create_page) && (type(g:wiki_map_create_page) == v:t_func || exists('*' . g:wiki_map_create_page))
         \ ? call(g:wiki_map_create_page, [a:page])
         \ : a:page
   call wiki#url#parse('wiki:/' . l:page).follow()


lervag:
I agree with the sentiment of being consistent, but when I look into this I believe the fix should go the other way. That is, I am pretty sure `g:wiki_map_create_page` can not refer to a funcref, because funcref variable names need to be capitalized. So, even with the proposed change, you could not do e.g. this:

```vim
function! Test()
  echo "hello world"
endfunction

let g:wiki_map_create_page = function('Test')
```

Because it would give this error message:

```
E704: Funcref variable name must start with a capital: g:wiki_map_create_page
```

Please correct me if I'm wrong. Else I think we should instead fix the other options and make it clear that they need to be strings, i.e. _names_ of the functions.

jgilchrist:
@lervag AIUI your comment is true for Vim but I'm just replicating https://github.com/lervag/wiki.vim/pull/174 as I'm also writing my configuration in Lua. I have:

```lua
function snakecase(text)
  return text:lower():gsub('%s+', '-')
end

vim.g.wiki_map_create_page = snakecase
vim.g.wiki_map_link_create = snakecase
```

So that if I start a wiki page called `My Wiki Page` I get `my_wiki_page.md`. This works for `wiki_map_link_create` because of #174 but `wiki_map_create_page` doesn't have the same logic.

jgilchrist:
Amazing! Thank you @lervag

Skabit:
> @lervag AIUI your comment is true for Vim but I'm just replicating #174 as I'm also writing my configuration in Lua. I have:
> 
> ```lua
> function snakecase(text)
>   return text:lower():gsub('%s+', '-')
> end
> 
> vim.g.wiki_map_create_page = snakecase
> vim.g.wiki_map_link_create = snakecase
> ```
> 
> So that if I start a wiki page called `My Wiki Page` I get `my_wiki_page.md`. This works for `wiki_map_link_create` because of #174 but `wiki_map_create_page` doesn't have the same logic.

@lervag I look the official documentation and I'm very lost. You wrote the next line:

`vim.g.wiki_map_create_page = function(x) return x:lower() end`

When I write that line in my intit file I have an error message when open neovim. I understand that you write the example of @jgilchrist in one code line or I'm forgetting something of lua syntaxis.

lervag:
@Skabit Please open a new issue and describe your problem in more detail. Please include the error message.

lervag:
(Btw, please note: I do not get any issues if I use the line from the docs in my config.)

PR #247, opened 2022-10-10
Add orgmode support

troiganto:
Hi! :wave: I've been recently been starting to get my notes more organized and was very impressed by your plugin! In particular, I liked that it took a more filetype-agnostic approach than other plugins, which focus much more on .wiki files specifically.

I've been trying to get a setup that uses both [nvim-orgmode](https://github.com/nvim-orgmode/orgmode) and wiki.vim. I quickly found that I couldn't _quite_ customize link creation from my vimrc file, so I had to modify the plugin itself.

If you're interested in maintaining support for orgmode files, I'd be delighted to upstream the changes I made. They  concern four corners:
- **link toggling**: Used to go wiki→md→wiki, now goes wiki→md→org→wiki;
- **tag parsing**: I've added `g:wiki#tags#orgmode_parser`, which is _not_ active by default; users have to opt into it by adding it to `g:wiki_tag_parsers`.
- **page renaming**: The simplest change, I just had to add a regex for orgmode-style links.
- **omnifunc**: Made it optional via `g:wiki_completion_enabled` because [nvim-orgmode](https://github.com/nvim-orgmode/orgmode) supplies its own omnifunc that would get unconditionally overridden by wiki.vim's.

I've tried to split the changes into reasonably-sized commits and added further explanations to each commit message. I think I followed your general style and added documentation where necessary, but please let me know if something's missing. I'd be happy to make further changes.

I've also added whatever unit tests I could think of, but wasn't extremely thorough. Let me know if there's another use case you'd like to see covered.

In any case, thank you again for the wonderful plugin and have a nice day!


lervag:
> Hi! I've been recently been starting to get my notes more organized and was very impressed by your plugin! In particular, I liked that it took a more filetype-agnostic approach than other plugins, which focus much more on .wiki files specifically.

Thanks, I'm glad you like it!

> If you're interested in maintaining support for orgmode files, I'd be delighted to upstream the changes I made.

I don't mind it at all! My first impression of this PR is that it is well written and thorough, and that it fits nicely with the core ideas of wiki.vim. Still, it is a large PR, so I'll want to review it properly. I'll do that as fast as I can.

> **link toggling**: Used to go wiki→md→wiki, now goes wiki→md→org→wiki;

Cool. I'm curious, though; should this be active by default? It would affect existing users who do use orgmode anyway. Perhaps it would be better to keep the current default toggle map and instead document how to update? I believe it could be more or less a single line of configuration.

> **omnifunc**: Made it optional via `g:wiki_completion_enabled` because [nvim-orgmode](https://github.com/nvim-orgmode/orgmode) supplies its own omnifunc that would get unconditionally overridden by wiki.vim's.

Perfectly reasonable.

> I've tried to split the changes into reasonably-sized commits and added further explanations to each commit message. I think I followed your general style and added documentation where necessary, but please let me know if something's missing. I'd be happy to make further changes.

I agree, this is well done.

> In any case, thank you again for the wonderful plugin and have a nice day!

Thank you for using it and for providing a PR for features you want! :)

Notice that I have a different PR incoming that makes large changes to the `journal.vim` and `date.vim` modules. I don't think it should conflict with any of your updates here, but you may be interested in checking: #249.

doc/wiki.txt
```diff
@@ -591,8 +603,11 @@ OPTIONS                                                   *wiki-config-options*
   Default: >
 
     let g:wiki_link_toggles = {
-          \ 'md': 'wiki#link#wiki#template',
           \ 'wiki': 'wiki#link#md#template',
+          \ 'md': 'wiki#link#org#template',
+          \ 'org': 'wiki#link#wiki#template',
+          \ 'adoc_xref_bracket': 'wiki#link#adoc_xref_inline#template',
+          \ 'adoc_xref_inline': 'wiki#link#adoc_xref_bracket#template',
```
lervag, line 610: Thanks, nice catch!

autoload/wiki/tags.vim
```diff
@@ -535,3 +535,36 @@ function! g:wiki#tags#default_parser.make(taglist, ...) dict abort
 endfunction
 
 " }}}1
+
+
+" {{{1 let g:wiki#tags#orgmode_parser = ...
+let g:wiki#tags#orgmode_parser = {
```
lervag, line 541: Would it make sense to reuse code from the default parser, i.e. like this:

```vim
let g:wiki#tags#orgmode_parser = deepcopy(g:wiki#tags#default_parser)
let g:wiki#tags#orgmode_parser.X = Y
```

lervag:
Good work; I've read through and I don't really have any major comments, except my question in the first reply and the question regarding possibility to avoid redundancy.

autoload/wiki/tags.vim
troiganto: I've given it a try and the length and number of regexes make it a bit difficult to fit this pattern within 79 columns. Let me know if the below looks good to you and I'll commit it:

```vim
" {{{1 let g:wiki#tags#orgmode_parser = ...
let g:wiki#tags#orgmode_parser = deepcopy(g:wiki#tags#default_parser)
call extend(g:wiki#tags#orgmode_parser, {
      \ 're_match': '\v^%(\c#\+filetags:\s+:[^: ]+:|\*.*:[^: ]+:$)',
      \ 're_findstart': '\v^%(^\c#\+filetags:\s+|\*.*)(:\zs[^: ]+)+$',
      \ 're_parse': '\v:\zs[^: ]+\ze:',
      \ 're_prefix': '\v^\zs.{-}\ze:([^: ]+:)+$'
      \})

function! g:wiki#tags#orgmode_parser.make(taglist, tagline) dict abort
  let l:prefix = matchstr(a:tagline, self.re_prefix)
  let l:tags = empty(a:taglist) ? '' : (':' . join(a:taglist, ':') . ':')
  return trim(l:prefix . l:tags)
endfunction

" }}}1
```

troiganto:
> Cool. I'm curious, though; should this be active by default? It would affect existing users who do use orgmode anyway. Perhaps it would be better to keep the current default toggle map and instead document how to update? I believe it could be more or less a single line of configuration.

That's a good question. I'll admit I don't quite understand the purpose of toggling between wiki and markdown links, since AFAIK one link type doesn't make sense in the other filetype.

That's not really relevant to this PR though. Maybe the following is a reasonable compromise?

```diff
 call wiki#init#option('wiki_link_toggles', {
       \ 'wiki': 'wiki#link#md#template',
       \ 'md': 'wiki#link#wiki#template',
+      \ 'org': 'wiki#link#org#template',
       \ 'adoc_xref_bracket': 'wiki#link#adoc_xref_inline#template',
       \ 'adoc_xref_inline': 'wiki#link#adoc_xref_bracket#template',
       \ 'date': 'wiki#link#wiki#template',
       \ 'shortcite': 'wiki#link#md#template',
       \ 'url': 'wiki#link#md#template',
       \})
```

That is: wiki/md links continue to toggle between each other as before. Orgmode links toggle onto themselves (i.e. don't do anything), but can be configured to do anything else.

This should not change people's workflows, but enabling orgmode only requires users to set `g:wiki_link_target_type` instead of also adding an entry to `g:wiki_link_toggles`.

(The alternative would be to add `wiki#link#org#matcher.toggle_template`, but that would make the links' behavior unconfigurable.)

Let me know what you think and I'll add a commit with the necessary changes. :D

autoload/wiki/tags.vim
lervag: Yes, that looks very good to me!

lervag:
Sorry for taking so long to reply!

> > Cool. I'm curious, though; should this be active by default? It would affect existing users who do use orgmode anyway. Perhaps it would be better to keep the current default toggle map and instead document how to update? I believe it could be more or less a single line of configuration.
> 
> That's a good question. I'll admit I don't quite understand the purpose of toggling between wiki and markdown links, since AFAIK one link type doesn't make sense in the other filetype.

I use this feature to change between wiki links and Markdown links - I tend to use Markdown links for everything external, and wiki links for internal links. I find it useful to be able to quickly toggle the different link styles.

> That's not really relevant to this PR though. Maybe the following is a reasonable compromise?

Yes, I think that is a good compromise, since it is backwards compatible while making it very easy to change one setting to get a more flexible link toggle chain.

troiganto:
No problem at all! I've made the relevant changes in additional commits and also rebased my branch to be on top of your changes to the journal code. Let me know if you find any other issues you'd like to see addressed :+1:

lervag:
Thanks! Merged now, I appreciate the contribution!

PR #277, opened 2023-03-09
Allow %Y to appear multiple times in a journal entry name

cwolfe3:
Only one byte!
I wanted journal entries to be in %Y/%m/%Y-%m-%d format so that if I ever flattened the hierarchy, I wouldn't have to rename any of the files. This broke many commands for the journal because I used %Y twice.


lervag:
Thanks!

PR #282, opened 2023-03-25
Lua: normalize paths in ui.select feature

trev-dev:
The new lua functions weren't working for me because I declared my root as `~/Wiki/`,  both the home relative path and trailing slash were tripping up the UI select and file picker.

Here's a fix for that.

### DIFF
--- a/lua/wiki/init.lua
+++ b/lua/wiki/init.lua
@@ -1,10 +1,12 @@
 local M = {}
 
+local function normalize_path(path)
+  return vim.fn.substitute(vim.fn.expand(vim.g.wiki_root .. "/"), "\\/$", "", nil)
+end
+
 -- Returns shorten path
 local function filter(path)
-  local root = vim.fn
-    .substitute(vim.fn
-      .expand(vim.g.wiki_root .. "/"), "\\/$", "", nil)
+  local root = normalize_path(path)
   return path:gsub(root, "")
 end
 
--- a/lua/wiki/init.lua
+++ b/lua/wiki/init.lua
@@ -2,7 +2,9 @@ local M = {}
 
 -- Returns shorten path
 local function filter(path)
-  local root = vim.g.wiki_root .. "/"
+  local root = vim.fn
+    .substitute(vim.fn
+      .expand(vim.g.wiki_root .. "/"), "\\/$", "", nil)
   return path:gsub(root, "")
 end
 


bybor:
Do you use Linux? Just curious how safe it is to get the latest `wiki.vim` code on Windows.

trev-dev:
> Do you use Linux? Just curious how safe it is to get the latest `wiki.vim` code on Windows.

Yep, I do.  Though, this feature was never Windows compatible to begin with.

lua/wiki/init.lua
```diff
@@ -1,8 +1,12 @@
 local M = {}
 
+local function normalize_path(path)
+  return vim.fn.substitute(vim.fn.expand(vim.g.wiki_root .. "/"), "\\/$", "", nil)
```
lervag, line 4: I believe you meant to pass `vim.g.wiki_root .. "/"` as the argument `path` here?

lervag: Also, I'm not sure if any of this is really needed. Notice how we already take care of this stuff [here](https://github.com/trev-dev/wiki.vim/blob/b4834fd66993a1286d1590b973f7f830295ec7f0/autoload/wiki/page.vim#L240-L262). I believe we can remove the `filter()` function and instead just use the output from `p:

https://github.com/trev-dev/wiki.vim/blob/b4834fd66993a1286d1590b973f7f830295ec7f0/lua/wiki/init.lua#L22-L23

lervag (changes requested):
Beside the bug, I believe my second comment should be further discussed here, and I would also be happy if @saccarosium  could participate since he was responsible for the initial lua code in #281.

I believe we can resolve this without using the filters at all. Or, at least for the `get_pages` function. Notice that `wiki#page#get_all` returns a list of pairs `(path, path_from_wiki_root)`, where `path_from_wiki_root` is the result of the filter itself. So, the question is how to make `vim.ui.select` use the second while displaying the list of pages and the first when opening it. Perhaps this is where we'll use the `format` option? Something like this:

```lua
local pages = vim.fn["wiki#page#get_all"]()
vim.ui.select(pages, {
  prompt = "WikiPages> ",
  format_item = function(pair) return pair[2] end
}, function(pair)
  vim.cmd("edit " .. pair[1])
end)
```

I think this may also work as expected on Windows, since `wiki#page#get_all` already does that work.

lua/wiki/init.lua
saccarosium: > Also, I'm not sure if any of this is really needed. Notice how we already take care of this stuff [here](https://github.com/trev-dev/wiki.vim/blob/b4834fd66993a1286d1590b973f7f830295ec7f0/autoload/wiki/page.vim#L240-L262). I believe we can remove the filter() function and instead just use the output from `p:

The filter function is used for simply making the full path relative to the wiki_path root. I don't really see why remove it.

saccarosium:
I've found the problem. I wasn't using the function `wiki#get_root()` that does all the normalizing all for you. To fix the bug we need to only apply this change.

```lua
-- Returns shorten path
local function filter(path)
    local root = vim.fn["wiki#get_root"]().. "/"
    return path:gsub(root, "")
end

-- Returns full path
local function unfilter(path)
    local root = vim.fn["wiki#get_root"]() .. "/"
    return root .. path
end
```

lua/wiki/init.lua
trev-dev: @saccarosium I think @lervag's built in already abstracts the relative path. Any qualms with just using what's there?

saccarosium: I wasn't aware if it built already something. If so probabbly better using those.

lervag:
> > Also, I'm not sure if any of this is really needed. Notice how we already take care of this stuff [here](https://github.com/trev-dev/wiki.vim/blob/b4834fd66993a1286d1590b973f7f830295ec7f0/autoload/wiki/page.vim#L240-L262). I believe we can remove the filter() function and instead just use the output from `p:
> 
> The filter function is used for simply making the full path relative to the wiki_path root. I don't really see why remove it.

No, the point was that the vimscript function already returns the path relative to the wiki root. The return value of `wiki#page#get_all()` is a list of path pairs where the first element is the absolute path and the second element is the path relative to the wiki root.

Thus, there is no reason to use the filter!

> I've found the problem. I wasn't using the function `wiki#get_root()` that does all the normalizing all for you. To fix the bug we need to only apply this change.
> 
> ```lua
> -- Returns shorten path
> local function filter(path)
>     local root = vim.fn["wiki#get_root"]().. "/"
>     return path:gsub(root, "")
> end
> 
> -- Returns full path
> local function unfilter(path)
>     local root = vim.fn["wiki#get_root"]() .. "/"
>     return root .. path
> end
> ```

Yes, `wiki#get_root()` seems a better thing to use here, but again, I think my previous suggestion is better. I'll just push it myself to show what I mean.

lervag:
I've pushed an update that I believe both 1) fixes the present issue, and 2) improves the code slightly by using the data already created by the vimscript apis.

Let me know what you think and if it works as expected on your ends as well.

trev-dev:
> I've pushed an update that I believe both 1) fixes the present issue, and 2) improves the code slightly by using the data already created by the vimscript apis.
> 
> Let me know what you think and if it works as expected on your ends as well.

Damn, too slow! Nice one :)

saccarosium:
> Let me know what you think and if it works as expected on your ends as well.

Work good on my end. The only little weird thing is that files are prececced by a `/`.

Screenshot 2023-03-25 at 22 11 44

lervag:
> Work good on my end. The only little weird thing is that files are prececced by a `/` (its only a visual thing).

Yes; I was not sure what would be the most appropriate solution here. It should be easy to remove the preceding slash. But do we want it? I guess one could claim it sort of indicates that the path is relative to wiki root?

lervag:
I think it looks better without the slash, so I pushed an update that removes it.

Issue #296, opened 2023-04-27
Making wiki.vim a multi-filetype wiki


saccarosium:
continues: #295

## Objective

Be able to simply work with multiple filetype and let `wiki.vim` change dynamically behaviour based on the user selected extensions.

## Solution

Having a global dictionary where define all the info we need for every supported filetype depending on the current filetype and selected filetypes of the user.

```vim
" global filetype dictionary for mainly use in the plugin but could also be modify by the user if needed
let g:wiki_filetypes = {
  \ 'markdown' : {
      \ 'link_type' : 'wiki'|'md',
      \ 'extension' : 'md | markdown',
      \ 'heading'   : 'atx | headex',
  \ },
  \ 'asciidoc' : {
     \ 'link_type' : 'wathever'
     \ 'extention' : 'adoc | asciidoc'
  \ },
  \ 'org' : {
     \ 'extension' : 'org'
  \ },
}
```

Dictionary where user can decide what filetype wiki.vim will consider as part of the wiki. The user needs to call the names of the filetypes, as defined in vim, because a filetype can have multiple extentions and we want this to work for every extention.

```vim
let g:wiki_active_filetypes = { markdown, asciidoc, text }
```

lervag:
Thanks! And sorry for replying late.

I think it would help to get some clear ideas of which concept we want to generalize for each filetype. Here are some I've recognized, perhaps there are more:

* Link type: The default link type to use when creating links.
* Extension: Which extension a URL should have.
* Default scheme: The default URL scheme when a url does not _have_ a scheme. Currently, this is hard coded to `wiki:`, but I think it should be generlaized. E.g. in `[[URL]]`, `URL` could have the `adoc` scheme if the user wanted.
* Heading: Which heading is used in the current markup. This is relevant for several things. Today, ATX is more or less generally assumed, except you helped generalize this somewhat for the TOC features.

The end goal would be an option similar to what you proposed, to configure behaviour for a given filetype with several available options. At first, I think perhaps the following would be quite easy to implement:

```vim
let g:wiki_filetype = {
      \ 'markdown': {
      \   'link_type': 'wiki'|'md',
      \   'link_extension': '.md',
      \   'index': 'index'|'README'|...,
      \ },
      \}
```

This would already deprecate the following options:

* `g:wiki_link_extension`
* `g:wiki_link_target_type`
* `g:wiki_index_name`

saccarosium:
[...]

PR #299, opened 2023-05-04
Support multi-filetype wikis with g:wiki_link_creation

lervag:
See the discussion in #296. This PR implements a first version of the proposed `g:wiki_link_creation`.

The new option provides a more flexible filetype-dependent behaviour. It deprecates four other options and makes configuring wiki.vim easier for most people. Since the new option provides sensible defaults, it also means most people will require less customization.

The following options are deprecated as the related features are all now _more flexibly_ controlled by `g:wiki_link_creation`:

* g:wiki_map_text_to_link
* g:wiki_map_create_page
* g:wiki_link_extension
* g:wiki_link_target_type

### DIFF
--- a/autoload/wiki/complete.vim
+++ b/autoload/wiki/complete.vim
@@ -119,7 +119,7 @@ function! s:completer_wikilink.complete_page(regex) dict abort " {{{2
 
   call map(l:cands, 'strpart(v:val, strlen(l:root)+1)')
   call map(l:cands,
-        \ empty(g:wiki_link_extension)
+        \ empty(wiki#link#get_creator('url_extension'))
         \ ? 'l:pre . fnamemodify(v:val, '':r'')'
         \ : 'l:pre . v:val')
   call s:filter_candidates(l:cands, a:regex)
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -64,6 +64,18 @@ endfunction
 
 "}}}1
 
+function! wiki#link#get_creator(...) abort " {{{1
+  let l:ft = expand('%:e')
+  if empty(l:ft) || index(g:wiki_filetypes, l:ft) < 0
+    let l:ft = g:wiki_filetypes[0]
+  endif
+  let l:c = get(g:wiki_link_creation, l:ft, g:wiki_link_creation._)
+
+  return a:0 > 0 ? l:c[a:1] : l:c
+endfunction
+
+" }}}1
+
 function! wiki#link#show(...) abort "{{{1
   let l:link = wiki#link#get()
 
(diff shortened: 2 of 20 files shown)


saccarosium (commented):
Seems to be all good. But, as I was testing, I've found a strange bug where I can modify values in `g:wiki_link_creation` because it says that I'm trying to index a nil value.

## INFO

* version: NVIM v0.9.0
* OS: MacOS Ventura
* Package manager: Paq (uses vim's default package management) 

What I've wrote:
```lua
-- lua
vim.g.wiki_link_creation.adoc.url_transform = function(x) return { x:lower():gsub(' ', '-'), x } end
```
```viml
" viml
let g:wiki_link_creation.adoc.url_transform = { x -> substitute(tolower(x), '\s\+', '-', 'g') }
```

_error_
```
Error detected while processing /Users/sacca/Repos/dot/config/nvim/init.lua:
E5113: Error while calling lua chunk: /Users/sacca/Repos/dot/config/nvim/init.lua:112: attempt to index field 'wiki_link_creation' (a nil value)
stack traceback:
        /Users/sacca/Repos/dot/config/nvim/init.lua:112: in main chunk
```

I've tried:
* calling variable from Lua with `vim.g` table
* using a viml file in `plugin` directory same problem
* using a viml file in `after/plugin` directory same problem
* calling viml with neovim `vim.cmd` function same problem
* changing the variable within command mode worked

lervag:
Ah, yes, I believe your problem is that you are trying to set a key of the option _before_ the option was fully defined. wiki.vim will define the default options when it is defined (it will not override user settings, though). But if you are trying to set options _before_ wiki.vim is loaded, then you must define the entire dictionary.

Since wiki.vim merges user options with the default, I believe the following should suffice:

```lua
vim.g.wiki_link_creation = {
  adoc = {
    url_transform = function(x) return x:lower():gsub(' ', '-') end
  }
}
```

```vim
let g:wiki_link_creation = {
      \ 'adoc': {
      \   'url_transform': { x -> substitute(tolower(x), '\s\+', '-', 'g') }
      \ }
      \}
```

I.e., the `url_extension` and `link_type` should be initialized by wiki.vim to the default values if you do not change them yourself.

lervag:
> Seems to be all good.

Oh, and thanks for testing! Do you have any thoughts on the choice of names?

> * calling variable from Lua with `vim.g` table
>
> * using a viml file in `after/plugin` directory same problem
>
> * calling viml with neovim `vim.cmd` function same problem

Hmm. I believe one or more of these _should_ work. It all depends on the load order, which you can inspect with `:scriptnames`. If you are trying to change only a key as in your example, then the `plugin/wiki.vim` file from wiki.vim must be sourced first. You could actually test this by adding `runtime plugin/wiki.vim` to your config file.

However, if you are _lazy loading_ wiki.vim, then this might not work. But then my previous suggestion of defining the entire dictionary should work.

saccarosium:
> Oh, and thanks for testing! Do you have any thoughts on the choice of names?

I think there are very clear. I like this changes a lot.

The only thing is figuring out this issue and document it. When I'm at home i will test your suggestions

lervag:
Thanks, I appreciate that! Perhaps we need to clarify something in the docs here.

PS! I will create some new issue with feature requests based on our other discussion. I'll try to make some "specs" and you if you want to take a shot at one of them, feel free. I advice that you let me know first, just to avoid that we work on the same thing in parallell.

saccarosium:
> ```lua
> vim.g.wiki_link_creation = {
>  adoc = {
>    url_transform = function(x) return x:lower():gsub(' ', '-') end
>  }
>}
>```

lervag:
I've now released v0.7 with an announcement. I'll therefore not mind merging these new changes. I believe this PR is ready to merge. Or did I miss anything? If I don't hear any objections I'll merge probably tomorrow.

saccarosium:
From my part I don't see any errors

PR #314, opened 2023-07-03
doc: fix example

jiz4oh:
the example is not corresponding with below and it's an invalid example, should be `md` instead


lervag:
Thanks, nice catch!

PR #318, opened 2023-07-25
Allowing the creation or modification of Wiki{Pages,Tag,Toc}

saccarosium:
Hi @lervag I'm back :),
When I first implemented the `vim.ui.select` (#281) backed for the `Wiki{Pages,Tag,Toc}`, I've overestimated the usefulness of this implementation. The function is very primitive and doesn't let you do anything that I will consider useful for searching my note (e.g. inability to open split from selected item and the lack of a preview).

For this reason I'm very confident very few are actually using the feature.

So since I think that adding every support for every fuzzy finder is not viable, and even if it was, we still could not represent the needs of all users, I've been thinking of an interface where people can overwrite/write the different backends for this commands. This can have the following benefits:
* Removing existing `wiki_fzf_pages_opts` and `wiki_fzf_tags_opts`, since the user will be able to simply overwrite the fzf backend;
* Make possible to keep this commands without creating new ones;
* Opens possibility for create plugins to add support for other backends and uis.

This implementation is very hacked toghether and very awkward to use both in lua and vimscript (need your infinite vimL knowledge to figure out a way to do it that makes sense).

### DIFF
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -74,7 +74,7 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer -nargs=+ -complete=customlist,wiki#complete#tag_names
         \ WikiTagRename call wiki#tags#rename_ask(<f-args>)
 
-  command! -buffer WikiToc call g:wiki_select_methods.toc()
+  command! -buffer WikiToc call g:wiki_select_method.toc()
   command! -buffer -nargs=1 WikiClearCache call wiki#cache#clear(<q-args>)
 
   if b:wiki.in_journal
--- a/autoload/wiki/buffer.vim
+++ b/autoload/wiki/buffer.vim
@@ -74,11 +74,7 @@ function! s:init_buffer_commands() abort " {{{1
   command! -buffer -nargs=+ -complete=customlist,wiki#complete#tag_names
         \ WikiTagRename call wiki#tags#rename_ask(<f-args>)
 
-  if has('nvim') && g:wiki_select_method == 'ui_select'
-    command! -buffer WikiToc lua require('wiki').toc()
-  else
-    command! -buffer WikiToc call wiki#fzf#toc()
-  endif
+  command! -buffer WikiToc call g:wiki_select_methods.toc()
   command! -buffer -nargs=1 WikiClearCache call wiki#cache#clear(<q-args>)
 
   if b:wiki.in_journal
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -706,119 +706,119 @@ The AsciiDoc link macro [0] is a generic link syntax that looks like this: >
 
   link:<target>[<attrlist>]
 
-The `[<attrlist>]` will be recognized by `wiki.vim` as the link text.
+  The `[<attrlist>]` will be recognized by `wiki.vim` as the link text.
 
-[0]: https://docs.asciidoctor.org/asciidoc/latest/macros/link-macro/
+  [0]: https://docs.asciidoctor.org/asciidoc/latest/macros/link-macro/
 
-Cite links ~
-                                                               *wiki-link-cite*
-The cite links are a convenient link type that can be used with e.g. Zotero
-URLs. That is, instead of writing a link with URL `zot:citekey` one may simply
-write `@citekey`.
+  Cite links ~
+  *wiki-link-cite*
+  The cite links are a convenient link type that can be used with e.g. Zotero
+  URLs. That is, instead of writing a link with URL `zot:citekey` one may simply
+  write `@citekey`.
 
-See also |wiki-link-url| for more info on the Zotero URL scheme and handler.
+  See also |wiki-link-url| for more info on the Zotero URL scheme and handler.
 
-Iso dates ~
-                                                               *wiki-link-date*
-Dates written in ISO date format, e.g. 2023-05-05, will be recognized as
-a link to `journal:2023-05-05`.
+  Iso dates ~
+  *wiki-link-date*
+  Dates written in ISO date format, e.g. 2023-05-05, will be recognized as
+  a link to `journal:2023-05-05`.
 
-==============================================================================
-OPTIONS                                                          *wiki-options*
+  ==============================================================================
+  OPTIONS                                                          *wiki-options*
 
-`wiki.vim` has several options to configure the behaviour. The following is
-a reference of all available options.
+  `wiki.vim` has several options to configure the behaviour. The following is
+  a reference of all available options.
 
-*g:wiki_cache_root*
+  *g:wiki_cache_root*
   Specify the cache directory for `wiki.vim`.
 
   Default value: OS dependent
 
-    `Linux and MacOS`: `'~/.cache/wiki.vim/'`
-    `Windows`:         Standard temporary folder (based on |tempname()|)
+  `Linux and MacOS`: `'~/.cache/wiki.vim/'`
+  `Windows`:         Standard temporary folder (based on |tempname()|)
 
-*g:wiki_cache_persistent*
+  *g:wiki_cache_persistent*
   Specify whether to use persistent caching.
 
   Default value: 1
 
-*g:wiki_completion_enabled*
+  *g:wiki_completion_enabled*
   Specify whether to override 'omnifunc' when entering a wiki buffer. Set this
   to 0 if you want to use another omnifunc.
 
   Default value: 1
 
-*g:wiki_completion_case_sensitive*
+  *g:wiki_completion_case_sensitive*
   Specify whether to use case-sensitive matching for omnicompletion of links
   and tags.
 
   Default value: 1
 
-*g:wiki_export*
+  *g:wiki_export*
   A dictionary that specifies the default configuration for |WikiExport|
   options. See the specifications for the options for more information.
 
   Default: >vim
 
-    let g:wiki_export = {
-          \ 'args' : '',
-          \ 'from_format' : 'markdown',
-          \ 'ext' : 'pdf',
-          \ 'link_ext_replace': v:false,
-          \ 'view' : v:false,
-          \ 'output': fnamemodify(tempname(), ':h'),
-          \}
+  let g:wiki_export = {
+    \ 'args' : '',
+    \ 'from_format' : 'markdown',
+    \ 'ext' : 'pdf',
+    \ 'link_ext_replace': v:false,
+    \ 'view' : v:false,
+    \ 'output': fnamemodify(tempname(), ':h'),
+    \}
 
-*g:wiki_filetypes*
-  List of filetype extensions for which `wiki.vim` should be enabled. Notice
-  that file extensions are not always the same as the filetype. For example,
-  the common extension for Markdown is `.md`, whereas the filetype is called
-  `markdown`. The option should list the file extensions.
+  *g:wiki_filetypes*
+    List of filetype extensions for which `wiki.vim` should be enabled. Notice
+    that file extensions are not always the same as the filetype. For example,
+         the common extension for Markdown is `.md`, whereas the filetype is called
+           `markdown`. The option should list the file extensions.
 
-  The first element of the list is considered the default filetype where that
-  is relevant.
+           The first element of the list is considered the default filetype where that
+           is relevant.
 
-  Default: `['md']`
+           Default: `['md']`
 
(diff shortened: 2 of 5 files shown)


lervag:
> Hi @lervag I'm back :)

Glad to see you! :)

> When I first implemented the `vim.ui.select` (#281) backed for the `Wiki{Pages,Tag,Toc}`, I've overestimated the usefulness of this implementation. The function is very primitive and doesn't let you do anything that I will consider useful for searching my note (e.g. inability to open split from selected item and the lack of a preview).
> 
> For this reason I'm very confident very few are actually using the feature.

Interesting observation, and I think you may be right.

> So since I think that adding every support for every fuzzy finder is not viable, and even if it was, we still could not represent the needs of all users, I've been thinking of an interface where people can overwrite/write the different backends for this commands.

That's the kind of interface I like!

> This can have the following benefits:
> 
> * Removing existing `wiki_fzf_pages_opts` and `wiki_fzf_tags_opts`, since the user will be able to simply overwrite the fzf backend;
> 
> * Make possible to keep this commands without creating new ones;
> 
> * Opens possibility for create plugins to add support for other backends and uis.

Yes, it would be nice to get some of these benefits.

> This implementation is very hacked toghether and very awkward to use both in lua and vimscript (need your infinite vimL knowledge to figure out a way to do it that makes sense).

I think you may be onto something here. But I'd like us to take one step back. I think it can be useful to consider how a user would want to customize things. In your PR, you did not provide any documentation for the new option. And I think we would also need to update the related docs for `:WikiPages` and similar.

I believe it can be easier to figure out how to implement things after we have thought about the user perspective.

This is your current proposal:

```vim
let g:wiki_select_methods = {
      \ 'fzf': {
      \   'pages': 'call wiki#wiki#fzf#pages()',
      \   'tags': 'call wiki#wiki#fzf#tags()',
      \   'toc': 'call wiki#fzf#toc()',
      \ },
      \ 'ui_select': {
      \   'pages': 'lua require("wiki.ui_select").pages()',
      \   'tags': 'lua require("wiki.ui_select").tags()',
      \   'toc': 'lua require("wiki.ui_select").toc()',
      \ },
      \}
```

And you would use `g:wiki_select_method` to specify the desired backend.

First: Perhaps it would be better to call the new option `g:wiki_select_backends`. Next, the new option is _only_ used together with the _new_ option. So, perhaps it is better to combine these.

As a first attempt, I propose this:

```vimhelp
*g:wiki_select_methods*
  A dictionary that specifies the selection method to be used for some
  commands. This allows the user quite a lot of flexibility in specifying which
  backend to use for which command, and also in creating custom backends as
  desired.

  The following table shows the option keys and the related commands:

    key     command
    ---     -------
    `pages` |WikiPages|
    `tags`  |WikiTags|
    `toc`   |WikiToc|

  The dictionary values can be one of two types:

    - |String|: The value must be either
        1) the name of a function to be executed, or
        2) a command line to be evaluated.
    - |Funcref|: A function that will be executed.

  The following builtin alternatives can be used:

    - fzf (https://github.com/junegunn/fzf.vim):
      - `wiki#fzf#pages()`
      - `wiki#fzf#tags()`
      - `wiki#fzf#toc()`
      - Note: This has a few additional options for more fine tuned control:
        - |g:wiki_fzf_pages_opts|
        - |g:wiki_fzf_tags_opts|
        - |g:wiki_fzf_pages_force_create_key|
    - Lua backend that uses |vim.ui.select| (neovim only!)
      - `require("wiki.ui_select").pages()`
      - `require("wiki.ui_select").tags()`
      - `require("wiki.ui_select").toc()`

  Default (neovim): >lua

    vim.g.wiki_select_methods = {
      pages = require("wiki.ui_select").pages,
      tags = 'equire("wiki.ui_select").tags,
      toc = require("wiki.ui_select").toc,
    }
<
  Default (Vim): >vim

    let g:wiki_select_methods = {
          \ 'pages': 'wiki#fzf#pages',
          \ 'tags': 'wiki#fzf#tags',
          \ 'toc': 'wiki#fzf#toc',
          \}
<
```

This would avoid adding a new option, but it would be a breaking change as we replace one option with a new one. I would also consider to remove the `g:wiki_fzf` options and instead somehow bake them into this new one. E.g., perhaps by using a lambda with an option block to the `fzf#...` functions as the default options.

What do you think?

lervag:
Also, note that I've recently implemented a better (IMHO) UI for input, confirm and select for other parts for neovim. I was not able to use `vim.ui.select` because the code is too tied to the standard pattern where the selection/input is blocking. But I think the new UI is quite good. Perhaps it would be possible to further improve it and allow it as one choice for this option as well..?

saccarosium:
> ```vim
> let g:wiki_select_methods = {
>      \ 'pages': 'wiki#fzf#pages',
>      \ 'tags': 'wiki#fzf#tags',
>      \ 'toc': 'wiki#fzf#toc',
>      \}
> ```
> This would avoid adding a new option, but it would be a breaking change as we replace one option with a new one. [...]

This makes definitely more sense. My idea is that the user could use either a `functionref` or a lambda function. I know how how this would looks like in lua, but I'm not sure how it is translated in viml:

```lua
-- As you show in the doc snippet
vim.g.wiki_select_methods = {
  pages = require("wiki.ui_select").pages,
  tags = require("wiki.ui_select").tags,
  toc = require("wiki.ui_select").toc,
}
-- or
vim.g.wiki_select_methods = {
  -- I don't know the telescope API is just an example
  pages = function() require'telescope'.find_files({cwd = vim.g.wiki_root}) end,
}
```

This is how I think of the final interface, at least in lua, for viml I don't really know. 
How do function in viml be passed to other functions (if you happen to know)? Do we only have the option of passing strings and then calling `call` on them?

> I would also consider to remove the g:wiki_fzf options and instead somehow bake them into this new one. E.g., perhaps by using a lambda with an option block to the fzf#... functions as the default options.

I don't really see value in this option since, I at least, would simply rewrite the function if there was something I didn't like. But probably better to keep it around since it doesn't hurt anybody. But also I think passing the parameter to a lambda would be a much nicer since you are not "polluting" the global vim "namespace" and could lead to a more descriptive and understandable experience (we could pass a dictionary with all flags you could activate `{ preview: "bat", layout: "reverce" }`).

> Also, note that I've recently implemented a better (IMHO) UI for input, confirm and select for other parts for neovim. I was not able to use vim.ui.select because the code is too tied to the standard pattern where the selection/input is blocking. But I think the new UI is quite good. Perhaps it would be possible to further improve it and allow it as one choice for this option as well..?

I see the new UI. I think is really neat. But I think that would be a good idea start thinking about dividing the plugin in separate plugin, perhaps the new UI is a good idea for a plugin named `wiki-ui.vim`, since is getting very big and I would love to see this plugin becoming more composable (but it is an other issue that should be further discussed).

lervag:
> This makes definitely more sense. My idea is that the user could use either a `functionref` or a lambda function. I know how how this would looks like in lua,

> ```lua
> -- As you show in the doc snippet
> vim.g.wiki_select_methods = {
>   pages = require("wiki.ui_select").pages,
>   tags = require("wiki.ui_select").tags,
>   toc = require("wiki.ui_select").toc,
> }
> -- or
> vim.g.wiki_select_methods = {
>   -- I don't know the telescope API is just an example
>   pages = function() require'telescope'.find_files({cwd = vim.g.wiki_root}) end,
> }
> ```

Yes, precisely. We can either specify a function directly or use a simple lambda to wrap a function call.

> This is how I think of the final interface, at least in lua, for viml I don't really know. How do function in viml be passed to other functions (if you happen to know)? Do we only have the option of passing strings and then calling `call` on them?

We can do similar things here. I am proposing three alternatives:

1. Using `Funcref`s (see `:help Funcref`) and lambdas.
2. Using the _name_ of a function and then the `call()` method (see `:help call()`).
3. Using a full command with `execute` similar to what you propose in your PR.

Both 2. and 3. means the value will be of type String. We can differ between them by checking with `exists('*function-name')`, see `:help exists()`). If the string is not recognized as a function name, then we will need to _assume_ it is a command.

> > I would also consider to remove the g:wiki_fzf options and instead somehow bake them into this new one. E.g., perhaps by using a lambda with an option block to the fzf#... functions as the default options.
> 
> I don't really see value in this option since, I at least, would simply rewrite the function if there was something I didn't like. But probably better to keep it around since it doesn't hurt anybody. But also I think passing the parameter to a lambda would be a much nicer since you are not "polluting" the global vim "namespace" and could lead to a more descriptive and understandable experience (we could pass a dictionary with all flags you could activate `{ preview: "bat", layout: "reverce" }`).

Yes, exactly, except - to avoid too much work - let's just use a string similar to what we do now.

```vim
" Alternative spec for default setting
let g:wiki_select_methods = {
      \ 'pages': { _ -> wiki#fzf#pages() },
      \}

" Add a previewer to the fzf options by passing an optional string argument
let g:wiki_select_methods = {
      \ 'pages': { _ -> wiki#fzf#pages('--preview "cat {1}"') },
      \}
```

This also shows how we can use a lambda function here in Vimscript, btw.

> > Also, note that I've recently implemented a better (IMHO) UI for input, confirm and select for other parts for neovim. I was not able to use vim.ui.select because the code is too tied to the standard pattern where the selection/input is blocking. But I think the new UI is quite good. Perhaps it would be possible to further improve it and allow it as one choice for this option as well..?
> 
> I see the new UI. I think is really neat. But I think that would be a good idea start thinking about dividing the plugin in separate plugin, perhaps the new UI is a good idea for a plugin named `wiki-ui.vim`, since is getting very big and I would love to see this plugin becoming more composable (but it is an other issue that should be further discussed).

Yes, I partly agree with this. With time, I've found I've implemented some "basic API functionality" that I would share between plugins. So, you will find things like these new UI features implemented _exactly_ the same in both wiki.vim and VimTeX. Similar for other features, e.g. the path API, the jobs API, and perhaps more. It is quite clear that there would be benefits to refactoring these parts _out_ into one (or more) separate library(-ies).

However, I also think there is a clear benefit to users of _avoiding_ dependencies. It puts some more work on my shoulders, but not much.

I think the most important thing for wiki.vim is that I (we) keep working on making it modular, robust and possibly also more composeable. But this may not necessarily mean that we should take things out; at least not without properly discussing and considering the costs and gains. If you feel strongly about this, please feel free to open a new issue to discuss it further.

saccarosium:
I've pushed some changes and documentation. At the end I think is best to keep using `g:wiki_select_method`, because to me makes more sense since at the end is only one function or another one and that how you call the function make more sense, at least to me.

If you like to keep it like this we should have a deprecation warning so people who where using the old interface can change to the new one (I didn't implement it yet)

lervag:
> I've pushed some changes and documentation. At the end I think is best to keep using `g:wiki_select_method`, because to me makes more sense since at the end is only one function or another one and that how you call the function make more sense, at least to me.

Yes, that's fine by me.

> If you like to keep it like this we should have a deprecation warning so people who where using the old interface can change to the new one (I didn't implement it yet)

Yes. Or, I'm not sure. I think perhaps it should suffice if we add a breaking change notice and then make sure to highlight the change in the next release notes. I would prefer not to maintain deprecation notices.

Now it seems two things remain:

* The Vimscript implementation.
* Adding some tests.

Do you want me to complete this, or do you want to work on it more first?

saccarosium:
Let work on it a little longer. Probably on Saturday I will finish it.

I want to:
- add the deprecation message
- add a telescope backend. Is the most used fuzzy finder on neovim, is probably a good idea to have it as an option (I'll keep `vim_ui_select` as the default tough)
- create a fzf-lua example to use it for myself and put it in the docs under `wiki-advanced-config-3`

Also is probably a good idea to put the select backend files in a subfolder in the lua directory if we want to extend the lua code (#319). Any suggestion for the name? I was thinking something like `lua/interfaces` or `lua/select`.

saccarosium:
I've finished most of the PR. The only things that are not done yet are:
- the telescope `toc` function is not working, I didn't found a way to get the selected item
- I've layed out option passing to the telescope functions but I don't pass anything to the calls

@lervag from now on you can finish the PR if you have the time.

lervag:
Thanks! Sorry for the delay, I've not had much time lately due to vacations and similar.

I've merged this now. I believe I fixed the issue you had with telescope for WikiToc. I also found a couple of minor bugs and I slightly improved the docs.

Feel free to open follow-up issues or PRs if I should have missed something!

lervag:
Sorry, I forgot to tag this PR while doing this. The merge commit is here: 0445962.

PR #336, opened 2024-02-11
Wrap fd path in quotes to preserve spaces

camfowler:
Prior to this, a space in the path would cause fd to interpret the
directory as two arguments, resulting in an error "is not a directory."
being the output of the completion.


lervag:
Thanks!

PR #350, opened 2024-04-26
Added template matching examples

sez11a:
Added template matching examples to the docs.

### DIFF
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -2182,6 +2182,28 @@ One may expect it to have one or more of the following keys:
     url and link text. The text part can be of interest if one wants to use it
     as the title of a document.
 
+Matching file names and directory structures is done a bit differently. To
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! This seems useful, but I think it could be better to place it under the example section for advanced configs. I.e. under `:help wiki-advanced-configuration`. We would need to update the introduction to be inclusive of more types of configuration, but I could do that after merging this PR.

In addition to moving the example, I would give it a title like "Example 4: Template matching" and a tag "wiki-advanced-config-4", and I would add a link to this tag under the help for `g:wiki_templates` and also under the `:help wiki-templates` introduction.

Does this make sense?

sez11a:
Absolutely; makes perfect sense.

lervag:
Are you still interested in bringing this in, or do you want me to finish it?

lervag:
I've finished this and merged it now.

PR #355, opened 2024-05-08
doc: add link.vim

qadzek:
Currently, I am a Vimwiki user, but I like the "do one thing and do it well" philosophy. I might give your plugin a try in the future, as the Vimwiki project seems a bit abandoned.

I developed [link.vim](https://github.com/qadzek/link.vim) for Vimwiki, but it might also be useful for `wiki.vim`, at least for users using Markdown syntax and Markdown links.

After briefly installing your plugin, everything seems to work as expected. `:help wiki-link-reference` also confirms that reference-style links are supported.

### DIFF
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -409,6 +409,11 @@ journal or diary in `wiki.vim`: >vim
           \ :<c-u>call wiki#journal#open()<cr>
   augroup END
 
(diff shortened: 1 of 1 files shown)


lervag:
Thanks! Your plugin seems interesting and useful.

PR #358, opened 2024-05-16
doc: fix key bindings to navigate journal

qadzek:
According to the help file, this should be `<C-p>` and `<C-n>`. I tried it, and that indeed seems to be the case.


lervag:
Thanks! <3

PR #359, opened 2024-05-16
doc: fix list of default mappings

qadzek:
It seems that two commands in the list of default mappings had an old name.
For another mapping, it looks like only Normal mode was documented.


lervag:
Thanks!

PR #366, opened 2024-06-16
Update nvim.vim for wiki#link#incoming_hover to work on nvim 0.7.2

ttdduu:
Hi lervag!

A long time ago I suggested a feature that you implemented in the `WikiLinkIncomingHover` function. Since I had actually forked your repo before the last release you made, I am just now starting to use it, and it is awesome now that I made it work on **nvim 0.7.2**.

I solved an error when **calling the function on a header that has incoming links** (which i can tell by calling `WikiClearCache links-in` and `WikiLinkIncomingToggle`):

```
Error detected while processing function wiki#link#incoming_hover[13]..wiki#ui#select[13]..wiki#ui#nvim#select[30]..wiki#ui#nvim#popup: line 49:
E5555: API call: Invalid key: 'buf'
Error detected while processing function wiki#link#incoming_hover[13]..wiki#ui#select: line 13:
E714: List required
```

After the changes, popup works as expected, but are you able to reproduce this error? nvim 0.7.2 is the newest version on Ubuntu's ppa, but it may still be older than the versions for which your original code applies.

### DIFF
--- a/autoload/wiki/ui/nvim.vim
+++ b/autoload/wiki/ui/nvim.vim
@@ -184,7 +184,7 @@ function! wiki#ui#nvim#popup(cfg) abort " {{{1
   " Create and fill the buffer
   let l:bufnr = nvim_create_buf(v:false, v:true)
   call nvim_buf_set_lines(l:bufnr, 0, -1, v:false, l:content)
-  call nvim_set_option_value('buftype', 'nofile', #{ buf: l:bufnr })
+  call nvim_buf_set_option(l:bufnr, 'buftype', 'nofile')
 
   " Create popup window
   let l:winopts = #{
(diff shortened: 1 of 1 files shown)


lervag:
> A long time ago I suggested a feature that you implemented in the `WikiLinkIncomingHover` function. Since I had actually forked your repo before the last release you made, I am just now starting to use it, and it is awesome now that I made it work on **nvim 0.7.2**.

Very glad to hear that!

> I solved an error when **calling the function on a header that has incoming links** (which i can tell by calling `WikiClearCache links-in` and `WikiLinkIncomingToggle`):

And sorry, but no, you did not really solve this error. The thing is, Neovim is evolving quite fast. Version 0.7.2 is _very_ old. And since wiki.vim version 0.9 I now depend on neovim version 0.9.5.

> nvim 0.7.2 is the newest version on Ubuntu's ppa, but it may still be older than the versions for which your original code applies.

I am sure there are easy ways to install more recent versions of Neovim. I don't know which approach is the best, but here are some alternatives:

* If you are on Ubuntu 24.04, then you should get version 0.9.5 immediately.
* Older Ubuntu, then maybe [this page](https://github.com/neovim/neovim/blob/master/INSTALL.md#ubuntu) has some hints.
* https://github.com/MordechaiHadad/bob
* https://mise.jdx.dev/ (this is actually an awesome tool that can install
different versions of very many useful tools)

----

So, I don't think it is correct to accept this PR, because it will break wiki.vim for everyone using more recent Neovim versions.

ttdduu:
I thought so! Thank you very much, I'll close this PR and update my nvim :)

lervag:
> I thought so! Thank you very much, I'll close this PR and update my nvim :)

You'll not regret it. Neovim is getting better all the time - no reason to stay on the old versions!

PR #370, opened 2024-06-27
doc: adjust WikiExport documentation

A-dot-S-dot:
Hey, 
it seems to be sufficient to set only `url_extension` to make `-link_ext_replace` work. It's my first pull request I ever made. So I appreciate every feedback.

### DIFF
--- a/doc/wiki.txt
+++ b/doc/wiki.txt
@@ -1960,8 +1960,8 @@ MAPPINGS AND COMMANDS REFERENCE                       *wiki-mappings-reference*
     in-browser wiki navigation, but for it to work well, it requires that:
 
       1. `-ext` is set to `html`, and
-      2. |g:wiki_link_creation| should have a `url_transform` that appends
-         the extension.
+      2. |g:wiki_link_creation| should have an entry with the key `html`,
+         which contains the option `url_extension` with the value `.html`.
 
   -output ~
     Set output directory where the exported file is stored. Relative paths are


doc/wiki.txt
```diff
@@ -1960,8 +1960,8 @@ MAPPINGS AND COMMANDS REFERENCE                       *wiki-mappings-reference*
     in-browser wiki navigation, but for it to work well, it requires that:
 
       1. `-ext` is set to `html`, and
-      2. |g:wiki_link_creation| should have a `url_transform` that appends
-         the extension.
+      2. |g:wiki_link_creation| should have an entry with the key `html`,
```
lervag, line 1963: Hmm. No, I don't think it's right that it needs an entry with the key `html`, thus I think this PR is wrong. However, we need that the entry that is selected from `g:wiki_link_creation` has the `url_transform` key.

lervag (commented):
> Hey, it seems to be sufficient to set only `url_extension` to make `-link_ext_replace` work. It's my first pull request I ever made. So I appreciate every feedback.

To be honest, it's a little bit unclear why you find the original phrasing to be wrong. Could you perhaps expand a little on what makes you think your suggested change is correct?

Considering the PR, it is clear that the phrasing must be confusing, and so I would be surprised if we can come up with an improvement here.

A-dot-S-dot:
Ok, then I had misunderstood the instructions. I thought you had to create a separate entry for html. Then it would be enough to use option `url_extension` for the extension. But I realize that I don't need this entry and that pandoc adjusts the links on its own. How was this requirement originally meant? Is it possible that it can simply be omitted?

lervag:
> Suppose you have the following code example? How I should adjust it regarding the doc?
> 
> …
> 
> With this settings I experience no issues. Is it possible that the requirement in the doc can simply be omitted?

Good question. Perhaps you are right. I'll look into it!

lervag:
I think you are right. I've updated the docs. I think the second requirement is not relevant, so I removed it.

PR #371, opened 2024-07-09
doc: Remove outdated text in g:wiki_templates

A-dot-S-dot:
After the example of `g:wiki_templates`  it is mentioned, that `;` is appended to the source filename. However,  I could not find `;` in the example above.  I assume this part of the doc is outdated. Thus, I suggest to remove this part in the doc.


lervag:
I think you are right. The "faulty" commit is here: 002e10b - it seems I forgot to remove that paragraph and noone noticed for three years (including me). Thanks!

PR #376, opened 2024-08-19
Improve fzf select

A-dot-S-dot:
Dear @lervag,
I suggest three improvements for `wiki#fzf#link` which also fixes a bug. I combined them in one pull request, because they depend on each other. 

1. The first commit removes `/` in the beginning of every search query in `wiki#fzf#pages` and `wiki#fzf#links`.
2. Unfortunately in the current state `wiki#fzf#link` does not work for new links, since the `mode` variable is incapsulated in existing pages. The second commit fixes this issue. 
3. The last commit adds a force key for `wiki#fzf#link` similar to the one of `wiki#fzf#pages`. It is worth mentioning that I changed `g:wiki_fzf_pages_force_create_key` to `g:wiki_fzf_force_create_key` to make it usable for both mentioned selecting methods. 

Best regards

### DIFF
--- a/autoload/wiki/fzf.vim
+++ b/autoload/wiki/fzf.vim
@@ -19,7 +19,7 @@ function! wiki#fzf#pages() abort "{{{1
   call fzf#run(fzf#wrap({
         \ 'source': map(
         \   wiki#page#get_all(),
-        \   {_, x -> x[0] . '#####' . x[1] }),
+        \   {_, x -> x[0] . '#####' . substitute(x[1], '^/', '', '') }),
         \ 'sink*': funcref('s:accept_page'),
         \ 'options': l:fzf_opts
         \}))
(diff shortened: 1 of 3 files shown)


lervag:
> 1. The first commit removes `/` in the beginning of every search query in `wiki#fzf#pages` and `wiki#fzf#links`.

Ok; this is for the _displayed_ field that you are searching in, right? The reason it was there is to make it clear that the page is "rooted". But I'm not sure if that's really necessary.

> 2. Unfortunately in the current state `wiki#fzf#link` does not work for new links, since the `mode` variable is incapsulated in existing pages. The second commit fixes this issue.

Ok, thanks!

> 3. The last commit adds a force key for `wiki#fzf#link` similar to the one of `wiki#fzf#pages`. It is worth mentioning that I changed `g:wiki_fzf_pages_force_create_key` to `g:wiki_fzf_force_create_key` to make it usable for both mentioned selecting methods.

Sounds useful, thanks!

I'll look at the code now and will probably merge this immediately.

PR #384, opened 2024-10-23
Add Markdown plugin render-markdown.nvim

qadzek:
This plugin is actively developed and has many powerful features. It is only available for Neovim though.


lervag:
Thanks!

PR #393, opened 2024-12-16
Remove mention of incorrect filename

qadzek:
If the `wiki.vim` configuration consists solely of `let g:wiki_root = '~/wiki'`, hitting `<leader>ww` actually opens `index.md` instead of `index.wiki`. This behavior changes when `let g:wiki_filetypes = [ 'wiki' ]` is added to the configuration.

In the similar section in `:help wiki-intro-usage` no filename is mentioned, so I've updated `README.md` accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ This outlines the basic steps to get started:
    let g:wiki_root = '~/wiki'
    ```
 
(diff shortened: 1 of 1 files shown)


lervag:
Thanks, good catch!

PR #422, opened 2025-12-25
Allow complete customization of setting link target

kceiw:
The problem I would like to solve is to create the link using the relative path instead of absolute path from `WikiLinkAdd`. It's on a markdown file and using md format. The file in the buffer and the link to add are in the same wiki root.

When I use `WikiLinkAdd` and select a file, the link target is always an absolute path from the wiki root, despite I try to customize it in vimrc file. 

In this change, the customized function will always be called to transform the link target, if the customized function exists. Then it'll fall back to the existing logic of using the absolute path.

### DIFF
--- a/autoload/wiki/link.vim
+++ b/autoload/wiki/link.vim
@@ -132,15 +132,17 @@ function! wiki#link#add(path, mode, ...) abort " {{{1
 
   let l:options = extend(l:defaults, a:0 > 0 ? a:1 : {})
 
-  if wiki#paths#is_abs(a:path)
+  let l:url = a:path
+  if has_key(l:creator, 'url_transform')
+    let l:transformed = l:creator.url_transform(a:path)
+    if !empty(l:transformed)
+      let l:url = l:transformed
+    endif
+  elseif wiki#paths#is_abs(a:path)
     let l:cwd = expand('%:p:h')
     let l:url = stridx(a:path, l:cwd) == 0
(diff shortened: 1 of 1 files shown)


lervag:
> The problem I would like to solve is to create the link using the relative path instead of absolute path from `WikiLinkAdd`. It's on a markdown file and using md format. The file in the buffer and the link to add are in the same wiki root.
> 
> When I use `WikiLinkAdd` and select a file, the link target is always an absolute path from the wiki root, despite I try to customize it in vimrc file.
> 
> In this change, the customized function will always be called to transform the link target, if the customized function exists. Then it'll fall back to the existing logic of using the absolute path.

Thanks, this does look like a safe and good change. It would be nice to have a couple of examples of what this change does. Like "input" -> "output" with and without the proposed change, given some relevant config.

lervag:
I'm still interested in the examples, but I find it hard to not agree to this change, so I'm merging regardless.

kceiw:
Sorry for the late reply and thanks for merging the change.

Here is an example. Note the `/` indicates from the wiki root.

1. I open this file `/Folder1/File1.md`.
1. I use `WikiLinkAdd`. It opens a file selector.
2. I select the file `/Folder2/File2.md`.

Before the change, it inserts this to `File1.md`: `[File2](/Folder2/File2.md)`. (I have a setting to use only file name for the link name. That is not relevant to this change).

With this change, it'll insert this to `File1.md`: `[File2](../Folder2/File2.md)`. (I have the setting for `url_transform`).

